PR #12, opened 2016-12-06
[GIRAPH-1125] Memory estimation mechanism for more efficient OOC execution

heslami:
**Thanks to Dionysios Logothetis for helping a lot with this diff.**

The new out-of-core mechanism is designed with the adaptivity goal in mind, meaning that we wanted the out-of-core mechanism to kick in only when it is necessary. In other words, when the amount of data (graph, messages, and mutations) all fit in memory, we want to take advantage of the entire memory. And, when in a stage the memory is short, only enough (minimal) amount of data goes out of core (to disk). This ensures a good performance for the out-of-core mechanism.

To satisfy the adaptiveness goal, we need to know how much memory is used at each point of time. The default out-of-core mechanism (ThresholdBasedOracle) get memory information based on JVM's internal methods (Runtime's freeMemory()). This method is inaccurate (and pessimistic), meaning that it does not account for garbage data that has not been purged by GC. Using JVM's default methods, OOC behaves pessimistically and move data out of core even if it is not necessary. For instance,
consider the case where there are a lot of garbage on the heap, but GC has not happened for a while. In this case, the default OOC pushes data on disk and immediately after a major GC it brings back the data to memory. This causes inefficiency in the default out of core mechanism. If out-of-core is used, but the data can entirely fit in memory, the job goes out of core even though going out of core is not necessary.

To address this issue, we need to have a mechanism to more accurately know how much of heap is filled with non-garbage data. Consequently, we need to change the Oracle (OOC policy) to take advantage of a more accurate memory usage estimation.

In this diff, we introduce a mechanism to estimate the amount of memory used by non-garbage data on the heap at each point of time. This estimation is based on the fact that Giraph is a data-parallel system in its essence, meaning that several types of threads exist, each type doing the same computation on various data. More specifically, we have compute/input threads, communication (receiving) threads, and OOC-IO threads. In a normal uniform execution, each type of threads behave
similarly and contribute similarly to each other on the memory footprint (meaning that different compute threads contribute similarly to each other on the memory footprint). In the proposed approach, we use a measure of progress for each type of thread and use linear regression to estimate the amount of memory.

The measure of progress for compute threads is the total number of vertices they have collectively processed in a superstep at each point, the measure of progress for communication threads is the total number of bytes received by a worker up to each point, and the measure of progress for IO threads is the amount of data read/written to disk up to each point during a superstep. These measures are restarted at the beginning of each superstep. We use these measures at the point where full GC
happens (when we have the accurate estimation of non-garbage data on the heap) and devise the linear model of used memory. We then use the linear model to estimate the amount of memory at each time based on the above progress measures.

### DIFF
--- a/giraph-core/src/main/java/org/apache/giraph/ooc/policy/MemoryEstimatorOracle.java
+++ b/giraph-core/src/main/java/org/apache/giraph/ooc/policy/MemoryEstimatorOracle.java
@@ -1,0 +1,834 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.apache.giraph.ooc.policy;
+
+import com.sun.management.GarbageCollectionNotificationInfo;
+import org.apache.commons.math.stat.regression.OLSMultipleLinearRegression;
+import org.apache.giraph.comm.NetworkMetrics;
+import org.apache.giraph.conf.FloatConfOption;
+import org.apache.giraph.conf.ImmutableClassesGiraphConfiguration;
+import org.apache.giraph.conf.LongConfOption;
+import org.apache.giraph.edge.AbstractEdgeStore;
+import org.apache.giraph.ooc.OutOfCoreEngine;
+import org.apache.giraph.ooc.command.IOCommand;
+import org.apache.giraph.ooc.command.LoadPartitionIOCommand;
+import org.apache.giraph.ooc.command.WaitIOCommand;
+import org.apache.giraph.worker.EdgeInputSplitsCallable;
+import org.apache.giraph.worker.VertexInputSplitsCallable;
+import org.apache.giraph.worker.WorkerProgress;
+import org.apache.log4j.Logger;
+
+import java.lang.management.ManagementFactory;
+import java.lang.management.MemoryPoolMXBean;
+import java.lang.management.MemoryUsage;
+import java.util.List;
+import java.util.Map;
+import java.util.Vector;
+import java.util.concurrent.atomic.AtomicLong;
+import java.util.concurrent.locks.Lock;
+import java.util.concurrent.locks.ReentrantLock;
+
+import static com.google.common.base.Preconditions.checkState;
+
+/**
+ * Implementation of {@link OutOfCoreOracle} that uses a linear regression model
+ * to estimate actual memory usage based on the current state of computation.
+ * The model takes into consideration 5 parameters:
+ *
+ * y = c0 + c1*x1 + c2*x2 + c3*x3 + c4*x4 + c5*x5
+ *
+ * y: memory usage
+ * x1: edges loaded
+ * x2: vertices loaded
+ * x3: vertices processed
+ * x4: bytes received due to messages
+ * x5: bytes loaded/stored from/to disk due to OOC.
+ *
+ */
+public class MemoryEstimatorOracle implements OutOfCoreOracle {
+  /** Memory check interval in msec */
+  public static final LongConfOption CHECK_MEMORY_INTERVAL =
+    new LongConfOption("giraph.garbageEstimator.checkMemoryInterval", 1000,
+      "The interval where memory checker thread wakes up and " +
+        "monitors memory footprint (in milliseconds)");
+  /** If mem-usage is above this threshold and no Full GC has been called,
+   * we call it manually. */
+  public static final FloatConfOption MANUAL_GC_MEMORY_PRESSURE =
+    new FloatConfOption("giraph.garbageEstimator.manualGCPressure", 0.95f, "");
+  /** Used to detect a high memory pressure situation */
+  public static final FloatConfOption GC_MINIMUM_RECLAIM_FRACTION =
+    new FloatConfOption("giraph.garbageEstimator.gcReclaimFraction", 0.05f, "");
+  /** If mem-usage is above this threshold, active threads are set to 0 */
+  public static final FloatConfOption AM_HIGH_THRESHOLD =
+    new FloatConfOption("giraph.amHighThreshold", 0.95f, "");
+  /** If mem-usage is below this threshold, active threads are set to max */
+  public static final FloatConfOption AM_LOW_THRESHOLD =
+    new FloatConfOption("giraph.amLowThreshold", 0.90f, "");
+  /** If mem-usage is above this threshold, creid is set to 0 */
+  public static final FloatConfOption CREDIT_HIGH_THRESHOLD =
+    new FloatConfOption("giraph.creditHighThreshold", 0.95f, "");
+  /** If mem-usage is below this threshold, credit is set to max */
+  public static final FloatConfOption CREDIT_LOW_THRESHOLD =
+    new FloatConfOption("giraph.creditLowThreshold", 0.90f, "");
+  /** OOC starts if mem-usage is above this treshold */
+  public static final FloatConfOption OOC_THRESHOLD =
+    new FloatConfOption("giraph.oocThreshold", 0.90f, "");
+
+  /** Logger */
+  private static final Logger LOG =
+    Logger.getLogger(MemoryEstimatorOracle.class);
+
+  /** Cached value for {@link #MANUAL_GC_MEMORY_PRESSURE} */
+  private final float manualGCMemoryPressure;
+  /** Cached value for {@link #GC_MINIMUM_RECLAIM_FRACTION} */
+  private final float gcReclaimFraction;
+  /** Cached value for {@link #AM_HIGH_THRESHOLD} */
+  private final float amHighThreshold;
+  /** Cached value for {@link #AM_LOW_THRESHOLD} */
+  private final float amLowThreshold;
+  /** Cached value for {@link #CREDIT_HIGH_THRESHOLD} */
+  private final float creditHighThreshold;
+  /** Cached value for {@link #CREDIT_LOW_THRESHOLD} */
+  private final float creditLowThreshold;
+  /** Cached value for {@link #OOC_THRESHOLD} */
+  private final float oocThreshold;
+
+  /** Reference to running OOC engine */
+  private final OutOfCoreEngine oocEngine;
+  /** Memory estimator instance */
+  private final MemoryEstimator memoryEstimator;
+  /** Keeps track of the number of bytes stored/loaded by OOC */
+  private final AtomicLong oocBytesInjected = new AtomicLong(0);
+  /** How many bytes to offload */
+  private final AtomicLong numBytesToOffload = new AtomicLong(0);
+  /** Current state of the OOC */
+  private volatile State state = State.STABLE;
+  /** Timestamp of the last major GC */
+  private volatile long lastMajorGCTime = 0;
+
+  /**
+   * Different states the OOC can be in.
+   */
+  private enum State {
+    /** No offloading */
+    STABLE,
+    /** Current offloading */
+    OFFLOADING,
+  }
+
+  /**
+   * Constructor.
+   * @param conf Configuration
+   * @param oocEngine OOC engine.:w
+   *
+   */
+  public MemoryEstimatorOracle(ImmutableClassesGiraphConfiguration conf,
+                               final OutOfCoreEngine oocEngine) {
+    this.oocEngine = oocEngine;
+    this.memoryEstimator = new MemoryEstimator(this.oocBytesInjected,
+      oocEngine.getNetworkMetrics());
+
+    this.manualGCMemoryPressure = MANUAL_GC_MEMORY_PRESSURE.get(conf);
+    this.gcReclaimFraction = GC_MINIMUM_RECLAIM_FRACTION.get(conf);
+    this.amHighThreshold = AM_HIGH_THRESHOLD.get(conf);
+    this.amLowThreshold = AM_LOW_THRESHOLD.get(conf);
+    this.creditHighThreshold = CREDIT_HIGH_THRESHOLD.get(conf);
+    this.creditLowThreshold = CREDIT_LOW_THRESHOLD.get(conf);
+    this.oocThreshold = OOC_THRESHOLD.get(conf);
+
+    final long checkMemoryInterval = CHECK_MEMORY_INTERVAL.get(conf);
+
+    Thread thread = new Thread(new Runnable() {
+      @Override
+      public void run() {
+        while (true) {
+          long oldGenUsageEstimate = memoryEstimator.getUsageEstimate();
+          MemoryUsage usage = getOldGenUsed();
+          if (oldGenUsageEstimate > 0) {
+            updateRates(oldGenUsageEstimate, usage.getMax());
+          } else {
+            long time = System.currentTimeMillis();
+            if (time - lastMajorGCTime >= 10000) {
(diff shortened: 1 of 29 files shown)


giraph-core/src/main/java/org/apache/giraph/utils/ThreadLocalProgressCounter.java
```diff
@@ -0,0 +1,67 @@
...
+        progress += entry.getValue();
+      }
+    }
+    return progress;
+  }
+
+  /**
+   * Removes all counters.
+   */
+  public void reset() {
```
edunov, line 64: What is the purpose of this function and how do you use it?

giraph-core/src/main/java/org/apache/giraph/ooc/policy/MemoryEstimatorOracle.java
```diff
@@ -0,0 +1,851 @@
...
+            return false;
+          }
+        }
+        if (equal(val2, 0)) {
+          return false;
+        }
+        factor.add(val1 / val2);
+      }
+
+      if (factor.size() < 2) {
```
edunov: This is the same as values.length < 2 and can be done in the beginning, right?

giraph-core/src/main/java/org/apache/giraph/ooc/policy/MemoryEstimatorOracle.java
```diff
@@ -0,0 +1,851 @@
...
+        return true;
+      }
+      double firstVal = factor.get(0);
+      for (int i = 1; i < factor.size(); ++i) {
```
edunov: Do you really need to create a Vector and this second loop? Can you just remember firstVal on line 812 and do the same check in the first loop?

giraph-core/src/main/java/org/apache/giraph/ooc/policy/MemoryEstimatorOracle.java
```diff
@@ -0,0 +1,851 @@
...
+      throws Exception {
+
+      if (coefficient.length != validColumnIndices.size()) {
+        throw new Exception("There are " + coefficient.length +
+          " coefficients, but " + validColumnIndices.size() +
+          " valid columns in the regression");
+      }
+
+      double[] beta = mlr.estimateRegressionParameters();
+      for (int i = 0; i < coefficient.length; ++i) {
```
edunov: nit: Arrays.fill(coefficient, 0);

giraph-core/src/main/java/org/apache/giraph/ooc/policy/MemoryEstimatorOracle.java
```diff
@@ -0,0 +1,851 @@
...
+        coefficient[i] = 0;
+      }
+      for (int i = 0; i < validColumnIndices.size(); ++i) {
+        coefficient[validColumnIndices.get(i)] = beta[i];
+      }
+      coefficient[5] = beta[validColumnIndices.size()];
```
edunov, line 790: Are you sure that bias is the last parameter? I did this small test: 

`   
   OLSMultipleLinearRegression mlr = new OLSMultipleLinearRegression();
    double[] y = new double[] {1, 2, 3};
    double[][] x = new double[][] {{0}, {0.1}, {0.2}};
    mlr.newSampleData(y, x);
    System.out.println(Arrays.toString(mlr.estimateRegressionParameters()));
`
And it printed: [0.9999999999999997, 10.0] , so bias is actually on the position 0

giraph-core/src/main/java/org/apache/giraph/ooc/policy/MemoryEstimatorOracle.java
```diff
@@ -0,0 +1,851 @@
...
+
+      // Create the coefficient table
+      boolean setIsValid = false;
+      lock.lock();
+      try {
+        if (validColumnIndices.size() >= 1 &&
+          dataSamples.size() >= validColumnIndices.size() + 1) {
+
+          double[][] xValues = new double[dataSamples.size()][];
+          double[] yValues = new double[memorySamples.size()];
```
edunov: I maybe miss something, but how do you guarantee that another thread will not add more values to memorySamples after you've created xValues but before you created yValues?  In other words, how do you guarantee that xValues and yValues have the same size?

giraph-core/src/main/java/org/apache/giraph/ooc/policy/MemoryEstimatorOracle.java
```diff
@@ -0,0 +1,851 @@
...
+   * based on linear regression.
+   *
+   * Upon a GC events, it gets updated with the most recent statistics through
+   * the {@link #addRecord} method.
+   */
+  private static class MemoryEstimator {
+    /** Stores the (x1,x2,...,x5) arrays of data samples, one for each sample */
+    private Vector<double[]> dataSamples = new Vector<>();
+    /** Stores the y memory usage dataSamples, one for each sample */
+    private Vector<Double> memorySamples = new Vector<>();
```
edunov: What is the point of using Vector everywhere in the class? Is it possible to use DoubleArrayList instead? I see a few pieces that can be simplified (e.g. conversion to primitive array)

giraph-core/src/main/java/org/apache/giraph/ooc/policy/MemoryEstimatorOracle.java
```diff
@@ -0,0 +1,851 @@
...
+          // After the computation of the regression, some coefficients may have
+          // values outside the valid value range. In this case, we set the
+          // coefficient to the minimum or maximum value allowed, and re-run the
+          // regression.
+          boolean changed;
+          extreme[3] = -1;
+          extreme[4] = -1;
+          do {
+            changed = refineCoefficient(4, 1, 2, xValues, yValues);
+            changed |= refineCoefficient(3, 0, 2, xValues, yValues);
```
edunov, line 657: What are these valid ranges? Please, add an explanation. 
oocBytes coefficient has to be between 1 and 2, why?
receivedBytes has to be between 0 and 2, again why?

giraph-core/src/main/java/org/apache/giraph/ooc/policy/MemoryEstimatorOracle.java
```diff
@@ -0,0 +1,851 @@
...
+
+        double value;
+        if (coefficient[coefIndex] < lowerBound) {
+          value = lowerBound;
+        } else {
+          value = upperBound;
+        }
+        int ptr = -1;
+        if (validColumnIndices.size() >= 1 &&
+          validColumnIndices.get(validColumnIndices.size() - 1) == coefIndex) {
```
edunov: This is pretty hardcore, can we have a loop instead? May loose a few CPU cycles, but it will be so much more maintainable

giraph-core/src/main/java/org/apache/giraph/ooc/policy/MemoryEstimatorOracle.java
```diff
@@ -0,0 +1,851 @@
...
+          }
+          // save new coefficient value in intermediate array
+          extreme[coefIndex] = value;
+          // re-run regression
+          mlr.newSampleData(yValues, xValues);
+          calculateRegression(coefficient, validColumnIndices, mlr);
+          result = true;
+        } else {
+          if (LOG.isDebugEnabled()) {
+            LOG.debug(
```
edunov, line 757: When and why do we get here? This branch doesn't seem to match javadoc for the method.

giraph-core/src/main/java/org/apache/giraph/ooc/policy/MemoryEstimatorOracle.java
```diff
@@ -0,0 +1,851 @@
...
+     *
+     * @param coefIndex Coefficient index
+     * @param lowerBound Lower bound
+     * @param upperBound Upper bound
+     * @param xValues double[][] matrix with data samples
+     * @param yValues double[] matrix with y samples
+     * @return True if coefficients were out-of-range
+     * @throws Exception
+     */
+    private boolean refineCoefficient(int coefIndex, double lowerBound,
```
edunov, line 712: This method is creepy :)  Why would any coefficient be out of range? Shouldn't we discard entire set of results in this case? And do you really improve results if you manually set the coefficient to lower or higher bound? 

One more thing to think about is how your approach makes all datapoints have the same weight. E.g. in the beginning of data load and in the end of it, the load might be very different. Maybe you can keep a sliding window of data? Say, for the last 10 minutes, instead of keeping everything since the beginning of the iteration. That will also limit the size of the data you run regression on (I have no idea what are the limits of OLS regressor in math4, but there has to be some limit since it is doing matrix decompositions)

edunov: Actually, is there a way to do regularization in this Java library? That would be a nice way to make sure you're coefficients stay within a reasonable range

giraph-core/src/main/java/org/apache/giraph/utils/ThreadLocalProgressCounter.java
heslami: This is not needed as this class is only used in `AbstractEdgeStore`, and consequently, is used in `MemoryEstimatorOracle`. Semantically, this method should be used to clear all the counters collected in one specific stage. To be accurate, every time the `MemoryEstimatorOracle#startIteration` is called, the reset should be called, as the estimation mechanism should be based on clean slate statistics in each iteration/stage. However, in our implementation of `MemoryEstimatorOracle` the values that remained constant throughout one stage are ignored in the estimation mechanism. And, `AbstractEdgeStore` is only used in the second phase of the input superstep. So, the values from the instance of `ThreadLocalProgressCounter` class are being ignored for the rest of the computation. That's why even if we do not reset the counter in `startIteration`, in our implementation, the program runs correctly.

Aside from all the description, I will add the call to `reset` in `startIteration` of the Oracle just to make the code more readable.

giraph-core/src/main/java/org/apache/giraph/ooc/policy/MemoryEstimatorOracle.java
heslami: Not really. `factors.size()` is the number of non-zero elements in the two columns of the given matrix. For instance, the following can be the matrix:
0, 0
1, 2
3, 5
0, 0
Here `values.length` is 4, but `factor.size()` is 2. Basically, the loop above this line weed out the rows that are 0 in both columns from the dependency check.

giraph-core/src/main/java/org/apache/giraph/ooc/policy/MemoryEstimatorOracle.java
heslami: Good point. Will change it 👍

giraph-core/src/main/java/org/apache/giraph/ooc/policy/MemoryEstimatorOracle.java
heslami: There is no explicit notion of "bias". "Bias" is what the user would add if he/she wants to. The returned parameters for your case should be an array of size 1. In other words, the setting you mentioned is equivalent to find "one" coefficient b where:
`[1 2 3] = b * [0 0.1 0.2]`
And the objective is to find the b where the sum of square root of errors is minimum.

When I run the exact same code as you mentioned I get an array of size one with the value "16.0" in it. Which makes sense.

The case you have had in mind can be expressed by the `x` array as follows:
`double[][] x = new double[][] {{0, 1}, {0.1, 1}, {0.2, 1}}`
or:
`double[][] x = new double[][] {{1, 0}, {1, 0.1}, {1, 0.2}}`
In the first case I get [10.000000000000002, 1.0], and in the second case I get [0.9999999999999997, 10.0]; and both make sense.

To put it more accurately, MLR can be applied in cases where you do not have any bias. If you want to have a bias, you should add "1" to all your `x` data (and the returned coefficient from MRL for this added "1" is your bias).

giraph-core/src/main/java/org/apache/giraph/ooc/policy/MemoryEstimatorOracle.java
heslami: The memory estimator is only changed in `addRecord` method, and this method is only called from `gcCompleted` which is a `synchronized` method. Other accesses to estimator is for reading from it (which is also guarded by the `lock` in the estimator object).

giraph-core/src/main/java/org/apache/giraph/ooc/policy/MemoryEstimatorOracle.java
heslami: `DoubleArrayList` from fastutil is useful when we have a lot of entry in the list. The number of entries in our case is very small (the number of full GCs in one superstep, which is in the range of 1-100, getting closer to 1000 in the very extreme cases). So, the choice of either `Vector` or `ArrayList` is more natural. I agree that we are not using the synchronized guarantees of `Vector`, hence using `ArrayList` is an even more natural choice. I'll change these to `ArrayList`. For the sake of simpler conversion from `ArrayList` to primitive array I'll use `DoubleArrayList` for the line you mentioned.

giraph-core/src/main/java/org/apache/giraph/ooc/policy/MemoryEstimatorOracle.java
heslami: Will add comments.

giraph-core/src/main/java/org/apache/giraph/ooc/policy/MemoryEstimatorOracle.java
heslami: Essentially, the `refineCoefficient` method is called in a loop to set the coefficients that are out-of-bound to be in their bound. This approach is a viable approach as our method of finding linear regression is based on least square method which finds the minimum of a convex function. This means, if a coefficient is out of bound in a non-constrained regression calculation, setting it to the extremum value of its bound results in the minimum of the convex function in our constrained regression problem.

Since this method is called in a loop, there is a chance that the method is called with a coefficient that is already refined and is not in `validColumnIndices` anymore. In such case, the coefficient is set to an extreme of its bound, but later on it will be overridden by the values in `extreme` array.

heslami: I'll add more comment for the `refineCoefficient` method to clarify what it does.

giraph-core/src/main/java/org/apache/giraph/ooc/policy/MemoryEstimatorOracle.java
heslami: A coefficient may fall out-of-range because of lack of enough data. Setting the coefficient to extremes of its range, actually, improves the accuracy. Note that with the regression we are trying to find a function to estimate the memory usage and having a reasonable coefficient for an index is better than entirely removing that index. This may result in a more pessimistic estimation, but 1) this will be transient and later on the coefficients be all correct and in their range naturally, and 2) this improves the accuracy of estimation.

As for the regularization, the memory estimation is reset once at the beginning of each superstep. The number of data points involved in the regression is at most the number of Full GCs in a particular superstep. In practice, the number of Full GCs in a superstep barely reaches to O(1000). As for the change in the load throughout the superstep, the mindset behind the regression is that it captures the changes. The assumption here is that we have multiple progress measures that each has the same pattern throughout one superstep. For instance, we assume that the computation has the same pattern in a supersptep -- we may experience slower or faster progress (in terms of number of vertices processed) throughout the superstep, but the computation itself is almost identical throughout the superstep. If we have higher load at the beginning of a superstep, the progress metrics are growing faster. Once the load drops towards the end of a superstep, the progress metrics grow at much lower rate. In general, the regression model should be agnostic to the load "in one superstep". With change of superstep, we may change a lot of things (messages, pattern of computation, etc) and that's why the estimation mechanism resets once a new superstep starts.

edunov (approved):
Looks good to me. Let's wait for results of testing this from Dio and then we can land it

dlogothetis:
Tests done for this diff:
-  Snapshots tests (those that were big enough to go out-of-core) both with the new and the old mechanism.
- Run 4 production jobs and verified that performance is better than the previous mechanism.

dlogothetis:
Somehow the file MemoryEstimatorOracle.java wasn't committed. I pulled the most recent changes, I can see the commit log (see below), but this file is missing.

commit f5b685efa09b539b1f95925405723f7ac7b1dcea
Author: Hassan Eslami <heslami@apache.org>
Date:   Fri Dec 23 12:03:37 2016 -0600

    GIRAPH-1125

    Closes #12

PR #15, opened 2017-01-13
Allow Pieces that don't go over vertices, but over "recepients of messages"

ikabiljo:
GIRAPH-1129

### DIFF
--- a/giraph-block-app-8/src/test/java/org/apache/giraph/block_app/framework/no_vtx/MessagesWithoutVerticesTest.java
+++ b/giraph-block-app-8/src/test/java/org/apache/giraph/block_app/framework/no_vtx/MessagesWithoutVerticesTest.java
@@ -1,0 +1,157 @@
+package org.apache.giraph.block_app.framework.no_vtx;
+
+import java.util.Iterator;
+
+import org.apache.giraph.block_app.framework.AbstractBlockFactory;
+import org.apache.giraph.block_app.framework.BlockUtils;
+import org.apache.giraph.block_app.framework.api.BlockWorkerReceiveApi;
+import org.apache.giraph.block_app.framework.api.BlockWorkerSendApi;
+import org.apache.giraph.block_app.framework.block.Block;
+import org.apache.giraph.block_app.framework.block.SequenceBlock;
+import org.apache.giraph.block_app.framework.piece.Piece;
+import org.apache.giraph.block_app.framework.piece.interfaces.VertexReceiver;
+import org.apache.giraph.block_app.framework.piece.interfaces.VertexSender;
+import org.apache.giraph.block_app.test_setup.TestGraphUtils;
+import org.apache.giraph.block_app.test_setup.graphs.EachVertexInit;
+import org.apache.giraph.block_app.test_setup.graphs.Small1GraphInit;
+import org.apache.giraph.conf.GiraphConfiguration;
+import org.apache.giraph.function.ObjectTransfer;
+import org.apache.giraph.graph.Vertex;
+import org.apache.giraph.types.ops.collections.array.WIntArrayList;
+import org.apache.hadoop.io.IntWritable;
+import org.apache.hadoop.io.NullWritable;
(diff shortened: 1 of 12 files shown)


giraph-block-app-8/src/test/java/org/apache/giraph/block_app/framework/no_vtx/MessagesWithoutVerticesTest.java
```diff
@@ -0,0 +1,157 @@
...
+    public Block createBlock(GiraphConfiguration conf) {
+      ObjectTransfer<Iterable<IntWritable>> msgsTransfer = new ObjectTransfer<>();
+      return new SequenceBlock(
+        new Piece<IntWritable, IntWritable, NullWritable, IntWritable, Object>() {
+          @Override
+          public VertexSender<IntWritable, IntWritable, NullWritable> getVertexSender(
+              BlockWorkerSendApi<IntWritable, IntWritable, NullWritable, IntWritable> workerApi,
+              Object executionStage) {
+            return (vtx) -> {
+              System.out.println("send: " + vtx);
```
edunov: Is there any need in having these? JUnit logs are already too noisy. Can we maybe have Asserts where it is appropriate?

giraph-block-app-8/src/test/java/org/apache/giraph/block_app/framework/no_vtx/MessagesWithoutVerticesTest.java
```diff
@@ -0,0 +1,157 @@
...
+  @Test
+  public void testWithGiraphEnv() throws Exception {
+    testSumOverSameGroup(3, true);
+  }
+
+  private void testSumOverSameGroup(int max, boolean fullGiraphEnv) throws Exception {
+    TestGraphUtils.runTest(
+        TestGraphUtils.chainModifiers(
+            new Small1GraphInit<IntWritable, IntWritable, NullWritable>(),
+            new EachVertexInit<>((vertex) -> vertex.getValue().set(- vertex.getId().get() % max))),
```
edunov: Just to confirm: this thing will also work if the message is sent to the existing vertex, correct? Can we have it in the test case as well?

giraph-block-app/src/main/java/org/apache/giraph/block_app/framework/piece/DefaultParentPiece.java
```diff
@@ -178,6 +178,14 @@ protected boolean allowOneMessageToManyIdsEncoding() {
     return false;
   }
 
+  /**
+   * Override to specify that receive of this Piece (and send of next Piece)
+   * ignore existing vertices, and just process received messages.
+   */
```
edunov, line 194: Would be nice to mention some potential applications or give examples.

giraph-block-app-8/src/test/java/org/apache/giraph/block_app/framework/no_vtx/MessagesWithoutVerticesTest.java
ikabiljo: yes - and 0 is such vertex in the test, though test might not fail in that case - will add.

giraph-block-app-8/src/test/java/org/apache/giraph/block_app/framework/no_vtx/MessagesWithoutVerticesTest.java
ikabiljo: ah, didn't know to avoid logs in tests - these were just for my initial debugging, asserts are above.

giraph-block-app/src/main/java/org/apache/giraph/block_app/framework/piece/DefaultParentPiece.java
ikabiljo: good idea, will do.

edunov (approved):

PR #16, opened 2017-01-20
Fix RepeatUntilBlock

ikabiljo:
https://issues.apache.org/jira/browse/GIRAPH-1130


edunov:
Igor, can you please explain what was broken? I don't see any description neither in jira ticket nor here

ikabiljo:
Added tests would be failing. 
When toQuit was passed true - no more pieces were executed within this block, i.e. toQuit was behaving as break in the loop after every piece, instead of being checked only when looping. Example:


```
ObjectHolder<Boolean> toQuit = new ObjectHolder<Boolean>();
RepeatUntilBlock.unlimited(
  new SequenceBlock(
    Pieces.masterCompute((master) -> toQuit.accept(true)),
    new Piece(),   <<-------- this piece wouldn't be executed, but should
  ),
  toQuit())
```

edunov (approved):

PR #21, opened 2017-03-01
GIRAPH-1132 Giraph jobs don't end if zookeeper dies before job starts

edunov:
I'm not sure I set all the timeouts right. There is no way to test all of these either. 
The idea is that we shouldn't have infinite wait loops anywhere. And that's exactly what this diff does

### DIFF
--- a/giraph-core/src/main/java/org/apache/giraph/conf/GiraphConstants.java
+++ b/giraph-core/src/main/java/org/apache/giraph/conf/GiraphConstants.java
@@ -88,6 +88,7 @@ import org.apache.hadoop.io.Writable;
 import org.apache.hadoop.io.WritableComparable;
 import org.apache.hadoop.mapreduce.OutputFormat;
 
+import static java.util.concurrent.TimeUnit.HOURS;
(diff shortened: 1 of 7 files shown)


giraph-core/src/main/java/org/apache/giraph/conf/GiraphConstants.java
```diff
@@ -1238,5 +1239,27 @@
   BooleanConfOption PREFER_IP_ADDRESSES =
       new BooleanConfOption("giraph.preferIP", false,
       "Prefer IP addresses instead of host names");
+
+  /**
+   * Timeout for "waitForever", when we need to wait for zookeeper.
+   * Since we should never really have to wait forever.
+   * We should only wait some reasonable but large amount of time.
+   */
+  LongConfOption WAIT_FOREVER_ZOOKEEPER_TIMEOUT_MSEC =
```
majakabiljo, line 1248: Nit: since we are not waiting forever anymore, I'd drop word forever from everywhere (forever and timeout have opposite meaning :-))

majakabiljo (approved):
👍

PR #24, opened 2017-03-07
GIRAPH-1134: Track number of input splits in command line

majakabiljo:
Summary: The progress we track during input reports how much data have we read, but not how much data there is to read.

Test Plan: Now it prints something like:
 Loading data: 70046603 vertices loaded, 658 vertex input splits loaded (out of 659); 7122062703 edges loaded, 4 edge input splits loaded (out of 1322)


edunov (approved):
LGTM!

PR #29, opened 2017-03-25
Expose block output conf options set

dlogothetis:
It's useful to be able to read whatever block output conf options have been set.


dlogothetis:
Tests:
- snapshot tests

edunov (approved):
+1

PR #30, opened 2017-03-27
[GIRAPH-1139] Fix resuming from checkpoint

neggert:
A couple of fixes that get resuming from checkpoint working.

* Set checkpointStatus to NONE in master when restarting from checkpoint.

Workers already do this, so the job hangs when restarting from checkpoint
while the master waits for workers to create checkpoints they're never
going to create.

* Set unique task id for each worker attempt

Previously, a worker would reuse the task id from the prior attempt. This
gets propagated to the Netty client id, which makes the master think it has
already processed any requests that come from that client, causing it to
discard them. This obviously causes problems.

And also a fix for GIRAPH-1136. We will now checkpoint on superstep 0 if checkpointing is enabled. Let me know if you'd rather I sent a separate PR for this.

Testing:
Ran custom Label Propagation implementation with checkpointing on a ~5b node graph. Manually killed workers (by logging in to worker node and running `kill -9 <pid>`. Ensured that Giraph successfully resumed from most recent checkpoint.

### DIFF
--- a/giraph-core/src/main/java/org/apache/giraph/bsp/BspService.java
+++ b/giraph-core/src/main/java/org/apache/giraph/bsp/BspService.java
@@ -188,12 +188,12 @@ public abstract class BspService<I extends WritableComparable,
   private long cachedApplicationAttempt = UNSET_APPLICATION_ATTEMPT;
   /** Job id, to ensure uniqueness */
   private final String jobId;
-  /** Task partition, to ensure uniqueness */
-  private final int taskPartition;
+  /** Task id, derived from partition and application attempt to ensure uniqueness */
+  private final int taskId;
   /** My hostname */
   private final String hostname;
-  /** Combination of hostname '_' partition (unique id) */
-  private final String hostnamePartitionId;
+  /** Combination of hostname '_' task (unique id) */
+  private final String hostnameTaskId;
   /** Graph partitioner */
   private final GraphPartitionerFactory<I, V, E> graphPartitionerFactory;
   /** Mapper that will do the graph computation */
@@ -231,8 +231,8 @@ public abstract class BspService<I extends WritableComparable,
     this.context = context;
     this.graphTaskManager = graphTaskManager;
     this.conf = graphTaskManager.getConf();
+
     this.jobId = conf.getJobId();
-    this.taskPartition = conf.getTaskPartition();
     this.restartedSuperstep = conf.getLong(
         GiraphConstants.RESTART_SUPERSTEP, UNSET_SUPERSTEP);
     try {
@@ -240,7 +240,6 @@ public abstract class BspService<I extends WritableComparable,
     } catch (UnknownHostException e) {
       throw new RuntimeException(e);
     }
-    this.hostnamePartitionId = hostname + "_" + getTaskPartition();
     this.graphPartitionerFactory = conf.createGraphPartitioner();
 
     basePath = ZooKeeperManager.getBasePath(conf) + BASE_DIR + "/" + jobId;
@@ -251,6 +250,8 @@ public abstract class BspService<I extends WritableComparable,
     inputSplitsAllDonePath = basePath + INPUT_SPLITS_ALL_DONE_NODE;
     applicationAttemptsPath = basePath + APPLICATION_ATTEMPTS_DIR;
     cleanedUpPath = basePath + CLEANED_UP_DIR;
+
+
 
     String restartJobId = RESTART_JOB_ID.get(conf);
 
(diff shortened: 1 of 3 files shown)


neggert:
Could someone please take a look at this? @majakabiljo @edunov

edunov:
Hi Nic, thank you for fixing this and sorry for the delay. 
I'll take a look at this diff other the weekend

giraph-core/src/main/java/org/apache/giraph/master/BspServiceMaster.java
```diff
@@ -1734,7 +1735,7 @@ private CheckpointStatus getCheckpointStatus(long superstep) {
     if (checkpointFrequency == 0) {
       return CheckpointStatus.NONE;
     }
-    long firstCheckpoint = INPUT_SUPERSTEP + 1 + checkpointFrequency;
+    long firstCheckpoint = INPUT_SUPERSTEP + 1;
```
edunov, line 1744: What is the reason for changing this? Do you want it to always do checkpoint after the first superstep?

edunov (commented):
@neggert, do you find checkpointing useful? Can you share more information about your use case? In our experience checkpointing adds more overhead than benefit

neggert:
I'm finding that roughly half of my ~2 hour job time is spent just loading the graph. Resuming from checkpoint lets me skip that step if something fails. This is also why I want to make sure that Giraph checkpoints before starting superstep 0. (It used to work this way, and it's documented in the Giraph book.)

giraph-core/src/main/java/org/apache/giraph/master/BspServiceMaster.java
neggert: This will actually checkpoint before running superstep 0. You're basically checkpointing the input loading work.

majakabiljo:
Can we get rid of getHostnamePartitionId() to avoid incorrectly using it in the future? I see various other places where taskPartition is used for identifier, do any of them need to be updated too?

neggert:
There actually is [one instance][1] where I think it's okay to use `getHostnamePartitionId`. This happens in `BspServiceMaster.becomeMaster` when creating a master bid in ZK, before any `TaskInfo` instance is created.

This does make me realize, though, that I need to make the same change to how the task id is set in `BspServiceMaster`.

What about just changing how `taskPartition` is set in `BspService`, like so?

    this.taskPartition = (int)getApplicationAttempt() * conf.getMaxWorkers() + getTaskPartition();

I think this is actually the minimal code change to fix the issue. I don't see anywhere in the code that actually cares about the task partition as anything other than a unique identifier.

[1]: https://github.com/apache/giraph/blob/trunk/giraph-core/src/main/java/org/apache/giraph/master/BspServiceMaster.java#L805

neggert:
This is ready for another look. I've replaced partition id with task id. The only place that actually needs a partition id is logging missing workers in `BspServiceMaster`. In that case, it's easy enough to recover the partition id by taking the task id modulo the number of workers.

neggert:
ping @edunov @majakabiljo

giraph-core/src/main/java/org/apache/giraph/bsp/BspService.java
```diff
@@ -272,7 +273,7 @@ public BspService(
     }
     if (LOG.isInfoEnabled()) {
       LOG.info("BspService: Connecting to ZooKeeper with job " + jobId +
-          ", " + getTaskPartition() + " on " + serverPortList);
+          ", " + getTaskId() + " on " + serverPortList);
```
majakabiljo: We need to set taskId before this

giraph-core/src/main/java/org/apache/giraph/bsp/BspService.java
```diff
@@ -288,6 +289,9 @@ public BspService(
       throw new RuntimeException(e);
     }
 
+    this.taskId = (int)getApplicationAttempt() * conf.getMaxWorkers() + conf.getTaskPartition();
```
majakabiljo: Please fix checkstyle (make sure 'mvn verify' passes)

majakabiljo (approved):
Thanks, +1, minor comments

giraph-core/src/main/java/org/apache/giraph/bsp/BspService.java
neggert: We can't actually set `taskId` before creating the ZK client, since we need ZK to get the get the application attempt. I changed this to log the partition instead, which we can get from `conf`.

neggert:
Fixed @majakabiljo's comments. @edunov

neggert:
Ping @edunov

edunov:
LGTM! Thanks @neggert 
@majakabiljo  do you want to land it or should I?

edunov:
I've committed this one. Let me know if there are any issues.

PR #32, opened 2017-03-30
GIRAPH-1140: Cleanup temp files in hdfs after job is done

majakabiljo:
Summary: Currently we are not cleaning up temp files we create in hdfs, fix it.

Test Plan: Ran a few jobs (successful, failed, killed), verified files are removed in all cases.


edunov (approved):
LGTM

PR #33, opened 2017-03-31
GIRAPH-1141: Kill the job if no progress is being made

majakabiljo:
Summary: Sometimes jobs can get stuck for various reasons, it's better to have an option to kill them then to keep them running holding resources.

Test Plan: Ran a large job with shorter limit and verified it gets killed. Also ran normal successful job. mvn verify

### DIFF
--- a/giraph-core/src/main/java/org/apache/giraph/job/CombinedWorkerProgress.java
+++ b/giraph-core/src/main/java/org/apache/giraph/job/CombinedWorkerProgress.java
@@ -150,10 +150,13 @@ public class CombinedWorkerProgress extends WorkerProgressStats {
     return workersDone == expectedWorkersDone;
   }
 
-  @Override
-  public String toString() {
+  /**
+   * Get string describing total job progress
+   *
+   * @return String describing total job progress
+   */
+  protected String getProgressString() {
     StringBuilder sb = new StringBuilder();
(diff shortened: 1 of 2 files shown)


giraph-core/src/main/java/org/apache/giraph/job/CombinedWorkerProgress.java
```diff
@@ -204,4 +215,19 @@ public String toString() {
...
     return sb.toString();
   }
+
+  /**
+   * Check if this instance made progress from another instance
+   *
+   * @param lastProgress Instance to compare with
+   * @return True iff progress was made
+   */
+  public boolean madeProgressFrom(CombinedWorkerProgress lastProgress) {
```
dlogothetis, line 225: Why not use the underlying raw numbers instead of the string? For instance, small changes in memory may not really mean progress.

majakabiljo: That's why I separated getProgressString from toString, to only contain actual progress. For different supersteps we are looking at different numbers so this seemed the easiest to compare instead of having all the if-s.

dlogothetis:
i missed this detail. looks ok to me.

edunov (approved):
LGTM

PR #34, opened 2017-04-07
Fixing MemoryEstimatorOracle

heslami:
This diff contains the following fixes:
1) Once the OOC is at OFFLAODING state, it may offload data to disk indefinitely. We should call GC manually and get OOC out of OFFLOADING if anything has changed in computation/communication pattern.
2) Fixes a "if" statement which entirely disabled regression for memory estimation.
3) Interleaving of in resetting and calculating memory estimation potentially can cause data race.
4) The superstep count in memory estimator was not coherent throughout the calculation.

"mvn clean install" passes all tests and checks.

### DIFF
--- a/giraph-core/src/main/java/org/apache/giraph/ooc/policy/MemoryEstimatorOracle.java
+++ b/giraph-core/src/main/java/org/apache/giraph/ooc/policy/MemoryEstimatorOracle.java
@@ -63,6 +63,10 @@ import static com.google.common.base.Preconditions.checkState;
  * x4: bytes received due to messages
  * x5: bytes loaded/stored from/to disk due to OOC.
  *
+ * If the estimation error is high, the oracle falls back to a threshold-based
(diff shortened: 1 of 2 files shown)


giraph-core/src/main/java/org/apache/giraph/ooc/policy/MemoryEstimatorOracle.java
```diff
@@ -231,58 +333,71 @@ public void run() {
...
       }
     } else {
+      lastNonHighPressureTimeMillis = System.currentTimeMillis();
       return new IOAction[]{IOAction.LOAD_PARTITION};
     }
   }
 
   @Override
   public boolean approve(IOCommand command) {
+    if (!(command instanceof WaitIOCommand) &&
```
dlogothetis, line 388: Can you add some documentation for the logic here?

PR #36, opened 2017-05-04
GIRAPH-1146: Keep track of number of supersteps when possible

majakabiljo:
Summary: In many cases we know how many supersteps are there going to be. We can keep track of it and log it with progress.

Test Plan: Ran a job, example log line:
 Data from 3 workers - Compute superstep 5 (out of 6): 171824 out of 1304814 vertices computed; 19 out of 252 partitions computed

### DIFF
--- a/giraph-block-app/src/main/java/org/apache/giraph/block_app/framework/BlockUtils.java
+++ b/giraph-block-app/src/main/java/org/apache/giraph/block_app/framework/BlockUtils.java
@@ -26,6 +26,7 @@ import org.apache.giraph.block_app.framework.api.giraph.BlockComputation;
 import org.apache.giraph.block_app.framework.api.giraph.BlockMasterCompute;
 import org.apache.giraph.block_app.framework.api.giraph.BlockWorkerContext;
 import org.apache.giraph.block_app.framework.block.Block;
+import org.apache.giraph.block_app.framework.block.PieceCount;
 import org.apache.giraph.block_app.framework.piece.AbstractPiece;
 import org.apache.giraph.block_app.framework.piece.Piece;
 import org.apache.giraph.conf.BooleanConfOption;
@@ -146,6 +147,11 @@ public class BlockUtils {
     Block executionBlock = blockFactory.createBlock(immConf);
     checkBlockTypes(
         executionBlock, blockFactory.createExecutionStage(immConf), immConf);
+
+    PieceCount pieceCount = executionBlock.getPieceCount();
+    if (pieceCount.isKnown()) {
+      GiraphConstants.SUPERSTEP_COUNT.set(conf, pieceCount.getCount() + 1);
+    }
 
     // check for non 'static final' fields in BlockFactories
     Class<?> bfClass = blockFactory.getClass();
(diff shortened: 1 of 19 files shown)


giraph-block-app/src/main/java/org/apache/giraph/block_app/framework/block/EmptyBlock.java
```diff
@@ -36,4 +36,9 @@
   @Override
   public void forAllPossiblePieces(Consumer<AbstractPiece> consumer) {
   }
+
+  @Override
+  public PieceCount getPieceCount() {
+    return new PieceCount(1);
```
ikabiljo: should be 0

giraph-block-app/src/main/java/org/apache/giraph/block_app/framework/block/PieceCount.java
```diff
@@ -0,0 +1,86 @@
...
+    return this;
+  }
+
+  public PieceCount multiply(int value) {
+    count *= value;
+    return this;
+  }
+
+  public int getCount() {
+    return known ? count : Integer.MAX_VALUE;
```
ikabiljo: this might easily lead to overflow if anything is done with this number.

You should either fatal (better), or return 1M or something here.

giraph-block-app/src/main/java/org/apache/giraph/block_app/framework/piece/delegate/DelegatePiece.java
```diff
@@ -249,6 +250,15 @@ public void forAllPossiblePieces(Consumer<AbstractPiece> consumer) {
     }
   }
 
+  @Override
+  public PieceCount getPieceCount() {
+    PieceCount ret = new PieceCount(0);
```
ikabiljo: this should be 1, this executes all pieces simultaneously.

giraph-block-app/src/main/java/org/apache/giraph/block_app/framework/BlockUtils.java
```diff
@@ -147,6 +148,11 @@ public static void initAndCheckConfig(GiraphConfiguration conf) {
     checkBlockTypes(
         executionBlock, blockFactory.createExecutionStage(immConf), immConf);
 
+    PieceCount pieceCount = executionBlock.getPieceCount();
+    if (pieceCount.isKnown()) {
+      GiraphConstants.SUPERSTEP_COUNT.set(conf, pieceCount.getCount());
```
ikabiljo: shouldn't it be pieceCount.getCount() + 1 ?

ikabiljo (changes requested):

giraph-block-app/src/main/java/org/apache/giraph/block_app/framework/BlockUtils.java
majakabiljo: There will be X+1 supersteps, but they are going to be supersteps 0..X. Actually I can make +1 here and then -1 in the logging part to keep it clear.

giraph-block-app/src/main/java/org/apache/giraph/block_app/framework/block/PieceCount.java
majakabiljo: Good point, I'll throw instead

giraph-block-app/src/main/java/org/apache/giraph/block_app/framework/piece/delegate/DelegatePiece.java
```diff
@@ -249,6 +250,11 @@ public void forAllPossiblePieces(Consumer<AbstractPiece> consumer) {
     }
   }
 
+  @Override
```
ikabiljo, line 253: you don't need to extend, default is 1 for any piece :)

ikabiljo (approved):
shipit, minor comment inline

PR #38, opened 2017-05-18
[GIRAPH-1147] Store timestamps when various fractions of input were done

majakabiljo:
Summary: In order to evaluate how read stragglers affect job performance, add a way to expose timestamps when various fractions of input were done reading through counters.

Test Plan: Ran a big job and verified counters are set correctly.

### DIFF
--- a/giraph-core/src/main/java/org/apache/giraph/master/input/MasterInputSplitsHandler.java
+++ b/giraph-core/src/main/java/org/apache/giraph/master/input/MasterInputSplitsHandler.java
@@ -20,10 +20,13 @@ package org.apache.giraph.master.input;
 
 import org.apache.giraph.comm.MasterClient;
 import org.apache.giraph.comm.requests.ReplyWithInputSplitRequest;
+import org.apache.giraph.conf.StrConfOption;
 import org.apache.giraph.io.GiraphInputFormat;
 import org.apache.giraph.io.InputType;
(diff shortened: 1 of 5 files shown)


dlogothetis:
As opposed to timestamps why not set the counters to the time passed between the different fractions? That's going to be easier to parse quickly.

giraph-core/src/main/java/org/apache/giraph/master/input/MasterInputSplitsHandler.java
```diff
@@ -56,16 +69,39 @@
...
+  private final Map<InputType, Integer> numSplitsPerType =
+      new EnumMap<>(InputType.class);
+  /** How many splits per type have been read so far */
+  private final Map<InputType, AtomicInteger> numSplitsReadPerType =
+      new EnumMap<>(InputType.class);
+  /**
+   * Store in counters timestamps when we finished reading
+   * these fractions of input
+   */
+  private final double[] doneFractionsToStoreInCoutners;
```
dlogothetis: Typo in field name.

majakabiljo:
Updated with comments and tested again

dlogothetis:
Looks ok to me.

Btw, did you squash the commits? You don't need to, they are squashed automatically when the pull request is merged. And we don't loose the diff between updates.

majakabiljo:
Ah didn't know, will do in the future

PR #39, opened 2017-05-30
[GIRAPH-1148] Connected components - make calculate sizes work with l…

majakabiljo:
…arge number of components

Summary: Currently if we have a graph with large number of connected components, calculating connected components sizes fails because reducer becomes too large. Use array of handles instead.

Test Plan: Successfully ran the job which was failing without this change

### DIFF
--- a/giraph-block-app-8/src/main/java/org/apache/giraph/block_app/library/prepare_graph/UndirectedConnectedComponents.java
+++ b/giraph-block-app-8/src/main/java/org/apache/giraph/block_app/library/prepare_graph/UndirectedConnectedComponents.java
@@ -352,10 +352,12 @@ public class UndirectedConnectedComponents {
     Pair<LongWritable, LongWritable> componentToReducePair = Pair.of(
         new LongWritable(), new LongWritable(1));
     LongWritable reusableLong = new LongWritable();
-    return Pieces.reduceAndBroadcast(
+    return Pieces.reduceAndBroadcastWithArrayOfHandles(
         "CalcConnectedComponentSizes",
+        3137, /* Just using some large prime number */
         new BasicMapReduce<>(
             LongTypeOps.INSTANCE, LongTypeOps.INSTANCE, SumReduce.LONG),
+        vertex -> getComponent.get(vertex).get(),
         (Vertex<LongWritable, V, Writable> vertex) -> {
           componentToReducePair.getLeft().set(getComponent.get(vertex).get());
           return componentToReducePair;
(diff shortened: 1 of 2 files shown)


giraph-block-app/src/main/java/org/apache/giraph/block_app/library/Pieces.java
```diff
@@ -320,6 +325,91 @@ public String toString() {
...
+
+      @Override
+      public void registerReducers(
+          final CreateReducersApi reduceApi, Object executionStage) {
+        reducers = new ArrayOfHandles.ArrayOfReducers<>(
+            numHandles,
+            new Supplier<ReducerHandle<S, R>>() {
+              @Override
+              public ReducerHandle<S, R> get() {
+                return reduceApi.createLocalReducer(reduceOp);
```
dlogothetis: This means that the same ReduceOperation instance is going to be shared across the different handles. Not entirely sure how the (de)serialization will work here.

dlogothetis: Actually, because you're using an ArrayOfHandles, this is going to be serialized-deserialized as a whole. This should be ok.

Though, this also assumes that reduce operations are stateless. In general, they are but there's nothing that prevents from writing a stateful one. Maybe add a comment about this.

majakabiljo: Good catch, it didn't occur to me. I'll fix it not to reuse the same ReduceOperation object.

giraph-block-app-8/src/main/java/org/apache/giraph/block_app/library/prepare_graph/UndirectedConnectedComponents.java
```diff
@@ -352,10 +352,15 @@ Block calculateConnectedComponentSizes(
...
-    return Pieces.reduceAndBroadcast(
-        "CalcConnectedComponentSizes",
+    // This reduce operation is stateless so we can use a single instance
+    BasicMapReduce<LongWritable, LongWritable, LongWritable> reduceOperation =
         new BasicMapReduce<>(
-            LongTypeOps.INSTANCE, LongTypeOps.INSTANCE, SumReduce.LONG),
+            LongTypeOps.INSTANCE, LongTypeOps.INSTANCE, SumReduce.LONG);
+    return Pieces.reduceAndBroadcastWithArrayOfHandles(
+        "CalcConnectedComponentSizes",
+        3137, /* Just using some large prime number */
```
dlogothetis, line 361: Should this be configurable?

majakabiljo: I can't come up with a reason why someone would want to change it. This can start having problems only at trillion components which wouldn't work for many other reasons, for tiny ones this few reducers won't add any overhead, and for larger ones which were currently working this is still improvement since reducers are processed on many machines now.

dlogothetis: Sounds good. Looks good then.

PR #40, opened 2017-05-31
Fix initialization of IdAndValueArrayEdges and IdAndNullArrayEdges

dlogothetis:
The initialize() method for these implementations does not reset the underlying data structure (array) just like in other implementations (e.g. HashMapEdges). This introduces bugs when the OutEdges implementation is re-used during input.

Test plan:
- new unit tests
- mvn clean install
- internal unit tests
- internal snapshot tests

### DIFF
--- a/giraph-core/src/main/java/org/apache/giraph/edge/IdAndNullArrayEdges.java
+++ b/giraph-core/src/main/java/org/apache/giraph/edge/IdAndNullArrayEdges.java
@@ -21,6 +21,7 @@ import java.io.DataInput;
 import java.io.DataOutput;
 import java.io.IOException;
 import java.util.Iterator;
+import java.util.NoSuchElementException;
 
 import org.apache.giraph.conf.ImmutableClassesGiraphConfigurable;
 import org.apache.giraph.conf.ImmutableClassesGiraphConfiguration;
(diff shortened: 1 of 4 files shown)


giraph-core/src/main/java/org/apache/giraph/edge/IdAndNullArrayEdges.java
```diff
@@ -75,7 +77,7 @@ public void initialize(Iterable<Edge<I, NullWritable>> edges) {
 
   @Override
   public void initialize(int capacity) {
-    neighbors.setCapacity(capacity);
+    neighbors = idTypeOps.createArrayList(capacity);
```
majakabiljo: Should we call clear + setCapacity instead, to reuse objects when data is kept serialized? (I guess that's the case when this problem popped up)

giraph-core/src/main/java/org/apache/giraph/edge/IdAndValueArrayEdges.java
```diff
@@ -78,8 +81,8 @@ public void initialize(Iterable<Edge<I, E>> edges) {
 
   @Override
   public void initialize(int capacity) {
-    neighborIds.setCapacity(capacity);
-    neighborEdgeValues.setCapacity(capacity);
+    neighborIds = idTypeOps.createArrayList(capacity);
+    neighborEdgeValues = edgeTypeOps.createArrayList(capacity);
```
majakabiljo: Same here

majakabiljo (approved):
+1

PR #43, opened 2017-08-23
Upgrade json dependency version

dlogothetis:
https://issues.apache.org/jira/browse/GIRAPH-1153

Tests
- unit tests
- internal snapshot tests
- tested with app that uses mapping input


aching:
+1 LGTM

PR #46, opened 2017-09-07
GIRAPH-1157

majakabiljo:
Allow implementations of JobProgressTrackerService to extend kill job behavior

Test Plan: Ran a job with additional behavior in kill method and verified it's executed


dlogothetis:
Just to be clear, here you want to allow custom behavior based on the type of the kill (i.e. the message), right?

Looks good too me.

PR #47, opened 2017-09-13
JIRA-1158

yukselakinci:
Giraph changes to support time estimation


giraph-core/src/main/java/org/apache/giraph/job/DefaultJobProgressTrackerService.java
```diff
@@ -129,6 +135,36 @@ public void run() {
...
 
   /**
+   * Determine if the job will finish in allotted time.
+   * @param conf Giraph configuration
+   * @param job Job
+   * @param progress Current superstep progress.
+   * @return true it the job can finish in allotted time, false otherwise.
+   */
+  protected boolean canFinishInTime(GiraphConfiguration conf, Job job,
+      SuperstepProgress progress) {
```
majakabiljo: Why not just give combinedWorkerProgress to this method?

giraph-core/src/main/java/org/apache/giraph/job/DefaultJobProgressTrackerService.java
```diff
@@ -129,6 +135,36 @@ public void run() {
...
+    try {
+      LOG.error("Killing the job because it won't complete in max " +
+              "allotted time: " +
+              GiraphConstants.MAX_ALLOWED_JOB_TIME_MS.get(conf) / 1000 + "s");
+      job.killJob();
+    } catch (IOException e) {
+      LOG.error(
+        "Failed to kill the job which won't complete in max allotted time: ",
+        e);
+    }
```
majakabiljo: Please rebase and use killJobWithMessage method

giraph-core/src/main/java/org/apache/giraph/job/CombinedWorkerProgress.java
```diff
@@ -144,6 +143,38 @@ public CombinedWorkerProgress(Iterable<WorkerProgress> workerProgresses,
...
+   */
+  public long getCurrentSuperstep() {
+    return currentSuperstep;
+  }
+
+  /**
+   * Get workers in superstep
+   * @return Workers in superstep.
+   */
+  public long getWorkerseInSuperstep() {
```
majakabiljo: Typo

giraph-core/src/main/java/org/apache/giraph/job/DefaultJobProgressTrackerService.java
```diff
@@ -129,6 +137,20 @@ public void run() {
   }
 
   /**
+   * Determine if the job will finish in allotted time
+   * @param conf Giraph configuration
+   * @param job Job
+   * @param workerCount Worker count
```
majakabiljo: Do you need workerCount here, should you use workersInSuperstep? If you need it, rename to startedWorkerCount or something, this way it sounds like how many workers job needs.

majakabiljo (commented):
Thanks, final comments

giraph-core/src/main/java/org/apache/giraph/job/DefaultJobProgressTrackerService.java
yukselakinci: this is not the started worker count, but all the worker counts that are requested by the job. My implementation in Digraph will only calculate the expected completion time if all workers are included in the CombinedWorkerCount to make the calculation accurate .

giraph-core/src/main/java/org/apache/giraph/job/CombinedWorkerProgress.java
yukselakinci: Thanks

majakabiljo (approved):
Lgtm, I'll commit it

PR #48, opened 2017-09-15
GIRAPH-1159

majakabiljo:
DefaultJobProgressTrackerService: Only kill the job if it's still running

Summary: In killJobWithMessage, we need to check if the job has completed before killing it.

Test Plan: Verified kill is not called on already completed job


dlogothetis:
+1

PR #52, opened 2017-10-09
GIRAPH-1163

yukselakinci:
Add exception parameter to logError API

### DIFF
--- a/giraph-core/src/main/java/org/apache/giraph/job/JobProgressTracker.java
+++ b/giraph-core/src/main/java/org/apache/giraph/job/JobProgressTracker.java
@@ -46,7 +46,7 @@ public interface JobProgressTracker {
    * Call this when you want to log an error line and exception
    * object from any mapper to command line
    *
-   * KryoWritableWrapper.tryConvertFromByteArray can be used to
+   * KryoWritableWrapper.convertFromByteArray can be used to
    * get exception object back
    *
    * @param logLine Line to log
--- a/giraph-core/src/main/java/org/apache/giraph/job/JobProgressTracker.java
+++ b/giraph-core/src/main/java/org/apache/giraph/job/JobProgressTracker.java
@@ -43,13 +43,14 @@ public interface JobProgressTracker {
   void logInfo(String logLine);
 
   /**
-   * Call this when you want to log an error line from any mapper to command
-   * line
+   * Call this when you want to log an error line and exception
+   * object from any mapper to command line
    *
    * @param logLine Line to log
+   * @param exByteArray Exception byte array
    */
   @ThriftMethod
-  void logError(String logLine);
+  void logError(String logLine, byte [] exByteArray);
 
   /**
    * Notify that job is failing
--- a/giraph-core/src/main/java/org/apache/giraph/job/JobProgressTracker.java
+++ b/giraph-core/src/main/java/org/apache/giraph/job/JobProgressTracker.java
@@ -46,6 +46,9 @@ public interface JobProgressTracker {
    * Call this when you want to log an error line and exception
    * object from any mapper to command line
    *
+   * KryoWritableWrapper.tryConvertFromByteArray can be used to
+   * get exception object back
+   *
    * @param logLine Line to log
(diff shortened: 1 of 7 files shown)


yukselakinci:
Abandoning this PR since exceptions will be passed through serialization

giraph-core/src/main/java/org/apache/giraph/job/JobProgressTracker.java
```diff
@@ -43,13 +43,14 @@
   void logInfo(String logLine);
 
   /**
-   * Call this when you want to log an error line from any mapper to command
-   * line
+   * Call this when you want to log an error line and exception
+   * object from any mapper to command line
    *
    * @param logLine Line to log
+   * @param exByteArray Exception byte array
```
majakabiljo: Nit: add a comment here that KryoWritableWrapper.tryConvertFromByteArray can be used to get exception object back. (if this was java 8 I'd suggest having another method which takes Exception parameter and have default implementation of this one calling it, but it's not :-( )

giraph-core/src/main/java/org/apache/giraph/writable/kryo/KryoWritableWrapper.java
```diff
@@ -120,4 +126,50 @@ public static Writable wrapIfNeeded(Object object) {
...
+   */
+  public static <T> byte [] tryConvertToByteArray(T object) {
+    byte [] arr = null;
+    try {
+      KryoWritableWrapper<T> wrapper =
+              new KryoWritableWrapper<>(object);
+      arr = WritableUtils.toByteArrayUnsafe(wrapper);
+      // Checkstyle exception due to unsafe conversion
+      // CHECKSTYLE: stop IllegalCatch
+    } catch (Exception e) {
```
majakabiljo: Is there some exception you saw because of which you are putting this, or just as safety? (if latter, it might be better not to do it in these utility methods but from place where you call them)

giraph-core/src/main/java/org/apache/giraph/writable/kryo/KryoWritableWrapper.java
```diff
@@ -120,4 +126,50 @@ public static Writable wrapIfNeeded(Object object) {
...
+   * @param arr byte array
+   * @param <T> type
+   * @return original object
+   */
+  public static <T> T tryConvertFromByteArray(byte [] arr) {
+    T result = null;
+    try {
+      KryoWritableWrapper<T> wrapper =
+              new KryoWritableWrapper<>();
+      WritableUtils.fromByteArrayUnsafe(
```
majakabiljo: Any reason you chose to use unsafe instead of simple writeToByteArray/readFromByteArray?

majakabiljo (commented):
What kind of testing did you do?

giraph-core/src/main/java/org/apache/giraph/job/JobProgressTracker.java
yukselakinci: Done.

giraph-core/src/main/java/org/apache/giraph/writable/kryo/KryoWritableWrapper.java
yukselakinci: The reason I added was because, the caller could chose any T values, and if the object that was initially converted to byte array is not related to T, then it can throw. But more importantly, there are other callers of this functionality and I don't want to have repeating try-catch blocks. Another option is to add a new utility class that would include the try-catch. Let me know if you think that would be better.

giraph-core/src/main/java/org/apache/giraph/writable/kryo/KryoWritableWrapper.java
yukselakinci: Agreed.

yukselakinci:
In addition to standard tests, I run HelloPageRank application but explicitly throwing a DigraphDataException.

majakabiljo (approved):

PR #55, opened 2018-01-10
GIRAPH-1166: Allow MasterObserver to get superstep aggregated metrics

majakabiljo:
Summary: Pass superstep AggregatedMetrics to MasterObserver, to be able to analyze eg stragglers in jobs.

### DIFF
--- a/giraph-core/src/main/java/org/apache/giraph/master/BspServiceMaster.java
+++ b/giraph-core/src/main/java/org/apache/giraph/master/BspServiceMaster.java
@@ -984,6 +984,9 @@ public class BspServiceMaster<I extends WritableComparable,
       } else {
         printAggregatedMetricsToHDFS(superstep, aggregatedMetrics);
       }
+      for (MasterObserver observer : observers) {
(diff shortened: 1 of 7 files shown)


giraph-core/src/main/java/org/apache/giraph/master/BspServiceMaster.java
```diff
@@ -984,6 +984,9 @@ private GlobalStats aggregateWorkerStats(long superstep) {
       } else {
         printAggregatedMetricsToHDFS(superstep, aggregatedMetrics);
       }
+      for (MasterObserver observer : observers) {
+        observer.superstepMetricsUpdate(superstep, aggregatedMetrics);
```
yukselakinci, line 988: Should we catch exceptions here?

yukselakinci (approved):

giraph-core/src/main/java/org/apache/giraph/master/BspServiceMaster.java
majakabiljo: Not sure, we don't do it on other actions of MasterObserver, implementation can always add safety inside of it.

PR #57, opened 2018-01-31
Instantiate OutEdges implementation with factory class

dlogothetis:
Sometimes the instantiation of an OutEdges implementation might have large overhead, e.g. if it access the configuration.  Instead of creating it directly, introduce a factory class that can be instantiated once. 

https://issues.apache.org/jira/browse/GIRAPH-1168

Testing:
- mvn clean install
- internal unit tests & snapshot tests
- run with large job with custom OutEdges factory.

### DIFF
--- a/giraph-core/src/main/java/org/apache/giraph/conf/GiraphConstants.java
+++ b/giraph-core/src/main/java/org/apache/giraph/conf/GiraphConstants.java
@@ -36,10 +36,12 @@ import org.apache.giraph.factories.ComputationFactory;
 import org.apache.giraph.factories.DefaultComputationFactory;
 import org.apache.giraph.factories.DefaultEdgeValueFactory;
 import org.apache.giraph.factories.DefaultMessageValueFactory;
+import org.apache.giraph.factories.DefaultOutEdgesFactory;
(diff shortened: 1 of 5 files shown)


yukselakinci (commented):
Did you check how much these changes improve performance? I was thinking that caching would be in the optimization path, but I don't see it. how does the factory class avoid unnecessary initializations? If number of calls to createOutEdges and createInputOutEdges is the same, it won't reduce the creation cost, right?

dlogothetis:
Good observation. Creating an OutEdges instance gets called a number of times in the order of the number of vertices. Now, this becomes an issue when the the creation of a new OutEdges instance does some expensive initialization, e.g. accessing parameters from the configuration can sometimes be expensive. With the factory, one can essentially cache this, upon the instantiation of the factory, and then pass the cached value to the OutEdges instance.

I've tried this with an application that does such an expensive operation adding mainly memory overhead, and the difference in performance is significant.

yukselakinci:
Are you not planning to implement caching as part of this check-in?

dlogothetis:
This is only introducing the factory that allows to do the caching, but how it's used for caching is application-specific, not part of this.

yukselakinci (approved):

PR #60, opened 2018-02-16
GIRAPH-1171: Collect stats about how long it took to process each partition

majakabiljo:
In order to make it easier to analyze whether there are some vertices in the graph which slow down the computation, or processing times of partitions is imbalanced, expose the stats about how long it took for each partition to be processed.


dlogothetis:
+1

PR #61, opened 2018-02-22
GIRAPH-1172 Removed dependency com.google.code.findbugs:annotations

jdasch:
Replaced `com.google.code.findbugs:annotations:2.0.2` with `com.github.stephenc.findbugs:findbugs-annotations:1.3.9-1` and `com.google.code.findbugs:jsr305:2.0.2`.

Had to remove a couple annotations that don't exist in the older version of the drop-in replacement.  There is a newer version of com.github.stephenc.findbugs:findbugs-annotations, but it hasn't been uploaded to maven central for some reason.  Moved usage of those annotations to the findbugs-exclude.xml file.

### DIFF
--- a/findbugs-exclude.xml
+++ b/findbugs-exclude.xml
@@ -96,4 +96,12 @@
       On the other hand, Kryo needs lambdas to be Serializable to work. -->
     <Bug pattern="SE_BAD_FIELD,SE_COMPARATOR_SHOULD_BE_SERIALIZABLE,SE_NO_SERIALVERSIONID" />
   </Match>
+  <Match>
+    <Class name="org.apache.giraph.block_app.library.gc.WorkerGCPiece"/>
+    <Bug pattern="DM_GC"/>
+  </Match>
+  <Match>
+    <Class name="org.apache.giraph.object.MultiSizedReusable"/>
+    <Bug pattern="SE_TRANSIENT_FIELD_NOT_RESTORED"/>
+  </Match>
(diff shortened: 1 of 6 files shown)


adamkennedy:
FindBugs is BSD, there's been some threads in other places indicating this is true based on the LICENSE file, but the maven package was built incorrectly.

jdasch:
I did come across this thread [1], which discusses that `com.google.code.findbugs:jsr305` is BSD, and documents that the JSR-305's BSD license is included in the findbugs repo [2].  But the license for FindBugs source is LGPL [3].

Unfortunately, Giraph is using `com.google.code.findbugs:annotations` which includes both JSR-305 and some FindBugs annotations [4], [5] which do have LGPL headers in the source listings.

[1] https://github.com/findbugsproject/findbugs/issues/128
[2] https://github.com/findbugsproject/findbugs/blob/2.0.2/findbugs/licenses/LICENSE-jsr305.txt
[3] https://github.com/findbugsproject/findbugs/blob/2.0.2/findbugs/licenses/LICENSE.txt
[4] https://github.com/findbugsproject/findbugs/blob/2.0.2/findbugs/src/java/edu/umd/cs/findbugs/annotations/SuppressFBWarnings.java
[5] https://github.com/findbugsproject/findbugs/blob/2.0.2/findbugs/src/java/edu/umd/cs/findbugs/annotations/SuppressWarnings.java

PR #63, opened 2018-03-13
Calculate edge count in log line only if necessary

dlogothetis:
Calling `partition.getEdgeCount()` iteration on all the vertices of the partition, which can be expensive. The expression inside `checkNotNull` is always evaluated, making this expensive. This constructs the string only if necessary.

https://issues.apache.org/jira/browse/GIRAPH-1175

Tests:
- mvn clean install
- run test jobs


majakabiljo (approved):
:-o

PR #65, opened 2018-03-22
GIRAPH-1181

yukselakinci:


giraph-core/src/main/java/org/apache/giraph/partition/ByteArrayPartition.java
```diff
@@ -233,14 +232,8 @@ public void write(DataOutput output) throws IOException {
       entry.getKey().write(output);
       // Note here that we are writing the size of the vertex data first
       // as it is encoded in the first four bytes of the byte[]
-      int vertexDataSize;
-      if (useUnsafeSerialization) {
```
majakabiljo: Are we not supporting the non-unsafe version anymore?

giraph-core/src/main/java/org/apache/giraph/utils/ExtendedInput.java
```diff
@@ -0,0 +1,326 @@
...
+import static org.apache.giraph.utils.ByteUtils.SIZE_OF_SHORT;
+import static org.apache.giraph.utils.ByteUtils.SIZE_OF_INT;
+import static org.apache.giraph.utils.ByteUtils.SIZE_OF_FLOAT;
+import static org.apache.giraph.utils.ByteUtils.SIZE_OF_DOUBLE;
+
+/**
+ * ExtendedInput allows kryo to read directly from the underlying stream
+ * without having to write to an interim buffer.
+ */
+public class ExtendedInput extends Input implements ExtendedDataInput {
```
majakabiljo: How does this differ from UnsafeByteArrayInputStream, I see a lot of code is being copied, can we reuse?

majakabiljo (commented):
Some high level comments first

giraph-core/src/main/java/org/apache/giraph/comm/aggregators/CountingOutputStream.java
```diff
@@ -36,7 +36,7 @@
    * Default constructor
    */
   public CountingOutputStream() {
-    dataOutput = new ExtendedByteArrayDataOutput();
+    dataOutput = new UnsafeByteArrayOutputStream();
```
majakabiljo, line 39: I see a lot of changes in aggregators related classes, why are we changing those in this diff?

giraph-core/src/main/java/org/apache/giraph/utils/UnsafeByteArrayOutputStream.java
```diff
@@ -36,7 +37,7 @@
  * Byte array output stream that uses Unsafe methods to serialize/deserialize
  * much faster
  */
-public class UnsafeByteArrayOutputStream extends OutputStream
+public class UnsafeByteArrayOutputStream extends Output
```
majakabiljo: I didn't realize kryo doesn't provide an interface but a class we have to extend. What kind of testing have you performed for this change, since it will affect everything we do? (testing both for performance and correctness)

giraph-core/src/main/java/org/apache/giraph/comm/aggregators/CountingOutputStream.java
yukselakinci: These changes allow us to write to the underlying buffer without having to first write to Chunked stream. However, what we write to stream is different depending on if the temporary Chunked streams is used, since the chunk size is also written. So we need to be consistent between write and read for kryo serialization. If the temporary chunk is not used when writing, casting the byte stream to ordinary ByteArrayInputStream when reading,  would cause error as we assume it used chunked stream. Since we know all our streams derives from UnsafeByteArray[Input/Outut]Stream, we can safely cast the byte array to unsafeByteArray stream.

giraph-core/src/main/java/org/apache/giraph/utils/UnsafeByteArrayOutputStream.java
yukselakinci: So far I run unit tests and snapshot tests and couple of prod jobs. I will run with more prod jobs to verify performance and correctness.

giraph-core/src/main/java/org/apache/giraph/utils/UnsafeReads.java
```diff
@@ -18,27 +18,24 @@
...
  * Byte array input stream that uses Unsafe methods to deserialize
  * much faster
  */
-public abstract class UnsafeReads implements ExtendedDataInput {
-
-  /** Buffer length */
-  protected int bufLength;
-  /** Position in the buffer */
-  protected long pos = 0;
+public abstract class UnsafeReads extends Input implements ExtendedDataInput {
```
dlogothetis, line 30: Can we add some unit tests for UnsafeArrayReads/UnsafeByteArrayInputStream just like with UnsafeByteArrayOutputStream?

dlogothetis:
While we're at this, maybe we can upgrade Kryo as well? Our version is 3 years old. If so, let's do this separately.

giraph-core/src/main/java/org/apache/giraph/utils/UnsafeByteArrayOutputStream.java
dlogothetis: I don't recall whether you've already done this, but can also do some microbenchmark, e.g. just write random objects and measure the difference?

yukselakinci: I have done some benchmarking with the additional changes that depend on this pull request ( I will send a separate PR for that). For this PR, my testing focused on correctness and making sure that there is no performance regression.

giraph-core/src/main/java/org/apache/giraph/utils/UnsafeReads.java
yukselakinci: TestUnsafeByteArrayOutputStream tests both input and output streams. I have added more testing in next PR through kryo serialization.

yukselakinci:
Good point, I will upgrade the kryo version separately.

giraph-core/src/main/java/org/apache/giraph/utils/ExtendedInput.java
yukselakinci: Done.

giraph-core/src/main/java/org/apache/giraph/partition/ByteArrayPartition.java
yukselakinci: I have added back the non-unsafe version

giraph-core/src/main/java/org/apache/giraph/utils/UnsafeByteArrayOutputStream.java
```diff
@@ -197,6 +201,13 @@ public void writeChar(int v) throws IOException {
     position += SIZE_OF_CHAR;
   }
 
+  @Override
+  public void writeChar(char v) {
```
dlogothetis: Just making sure, there are no other methods that are overloaded other than this one?

yukselakinci:
UnsafeByteArray streams now extends kryo Output/Input streams so that kryo serialization can directly write to this stream without using an additional buffer, providing a faster serialization.

Tests performed:
- Giraph unit tests
- Snapshot tests
- Run several workloads with different applications and measured correctness/performance, running time stayed in the boundary of the current values.

dlogothetis:
+1

PR #69, opened 2018-04-06
GIRAPH-1185

yukselakinci:
Second PR for serialization. There is one more.

### DIFF
--- a/giraph-core/src/main/java/org/apache/giraph/writable/kryo/HadoopKryo.java
+++ b/giraph-core/src/main/java/org/apache/giraph/writable/kryo/HadoopKryo.java
@@ -26,11 +26,15 @@ import java.util.Map;
 import java.util.Map.Entry;
 import java.util.Random;
 
+import com.esotericsoftware.kryo.util.DefaultClassResolver;
 import org.apache.giraph.conf.GiraphConfigurationSettable;
+import com.esotericsoftware.kryo.ClassResolver;
+import com.esotericsoftware.kryo.ReferenceResolver;
+import com.esotericsoftware.kryo.util.MapReferenceResolver;
 import org.apache.giraph.types.ops.collections.Basic2ObjectMap;
 import org.apache.giraph.types.ops.collections.BasicSet;
-import org.apache.giraph.writable.kryo.markers.KryoIgnoreWritable;
 import org.apache.giraph.writable.kryo.markers.NonKryoWritable;
+import org.apache.giraph.writable.kryo.markers.KryoIgnoreWritable;
 import org.apache.giraph.writable.kryo.serializers.ArraysAsListSerializer;
 import org.apache.giraph.writable.kryo.serializers.CollectionsNCopiesSerializer;
 import org.apache.giraph.writable.kryo.serializers.DirectWritableSerializer;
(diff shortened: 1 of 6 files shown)


giraph-core/src/main/java/org/apache/giraph/writable/kryo/KryoSimpleWrapper.java
```diff
@@ -0,0 +1,93 @@
...
+ *
+ * Unlike KryoWritableWrapper, this wrapper does not
+ * support recursive/nested objects.
+ *
+ * Optimized for performance,
+ *
+ *
+ * @param <T> Object type
+ */
+public class KryoSimpleWrapper<T> implements Writable {
```
majakabiljo: What's the difference between this and KryoWritableWrapper, when should we use one and when the other?

giraph-core/src/main/java/org/apache/giraph/writable/kryo/KryoSimpleWritable.java
```diff
@@ -0,0 +1,57 @@
...
+import org.apache.giraph.writable.kryo.markers.KryoIgnoreWritable;
+
+/**
+ * Class which you can extend to get all serialization/deserialization
+ * done automagically.
+ *
+ * This serialization does not support recursive/nested structures
+ * to provide faster serialization.
+ */
+public abstract class KryoSimpleWritable implements KryoIgnoreWritable {
```
majakabiljo: It would be great to switch to Java 8, to be able to keep this an interface :-)

giraph-core/src/main/java/org/apache/giraph/writable/kryo/HadoopKryo.java
```diff
@@ -111,19 +123,15 @@ public Kryo create() {
         "Logger must be a static field");
   }
 
-  // Use chunked streams, so within same stream we can use both kryo and
-  // non-kryo serialization.
   /** Reusable Input object */
-  private final InputChunked input = new InputChunked(4096);
+  private InputChunked input;
```
majakabiljo, line 129: Should we separate two Kryo implementations in two classes since they have a lot of differences?

giraph-core/src/main/java/org/apache/giraph/writable/kryo/HadoopKryo.java
```diff
@@ -203,15 +277,33 @@ public T execute(Kryo kryo) {
     });
   }
 
+  /**
+   * Returns a kryo which doesn't track objects, hence
+   * no recursive/nested objects should be serialized.
+   * @return Hadoop kryo which doesn't track objects.
+   */
+  public static HadoopKryo getNontrackingKryo() {
```
majakabiljo: How will someone using kryo know which implementation to use? Can we add some guidance around that?

giraph-core/src/main/java/org/apache/giraph/writable/kryo/KryoSimpleWrapper.java
yukselakinci: I have added more documentation now. The usage of KryoSimpleWrapper is similar to KryoWritabeWrapper. KryoSimpleWrapper can be used if the object being serialized is not recursive or nested, because it uses a version of kryo object that doesn't track references, hence significantly faster.

giraph-core/src/main/java/org/apache/giraph/writable/kryo/KryoSimpleWritable.java
yukselakinci: I agree :-)

giraph-core/src/main/java/org/apache/giraph/writable/kryo/HadoopKryo.java
yukselakinci: Agreed, we can refactor the existing code to have a better code readability. I have added a TODO for that.

giraph-core/src/main/java/org/apache/giraph/writable/kryo/HadoopKryo.java
yukselakinci: Added more documentation.

majakabiljo (approved):
Discussed offline, @yukselakinci will in a separate diff refactor HadoopKryo.

PR #70, opened 2018-04-17
GIRAPH-1188

yukselakinci:


giraph-core/src/main/java/org/apache/giraph/writable/kryo/GiraphClassResolver.java
```diff
@@ -0,0 +1,308 @@
...
+    }
+    memoizedClassId = classID;
+    memoizedClassIdValue = registration;
+    return registration;
+  }
+
+  /**
+   * Reset the internal state
+   */
+  public void reset() {
```
majakabiljo: Can you please add a comment for what this and the two methods below are for and why we don't have to implement them?

giraph-core/src/main/java/org/apache/giraph/writable/kryo/GiraphClassResolver.java
```diff
@@ -0,0 +1,308 @@
...
+   * explicity registered class IDs.
+   * */
+  private static final int BASE_CLASS_ID = 10000;
+
+  /** Class name to ID cache */
+  private static Map<String, Integer> CLASS_NAME_TO_ID = new HashMap();
+  /** ID to class name cache */
+  private static Map<Integer, String> ID_TO_CLASS_NAME = new HashMap();
+  /** Zookeeper */
+  private static ZooKeeperExt ZK;
```
majakabiljo, line 73: It's generally not a nice design to have these static, but not sure if because of some details in kryo there is no other way?

giraph-core/src/main/java/org/apache/giraph/writable/kryo/GiraphClassResolver.java
```diff
@@ -0,0 +1,308 @@
...
+import static com.esotericsoftware.kryo.util.Util.getWrapperClass;
+
+/**
+ * This class resolver assigns unique classIds for every class that was not
+ * explicitly registered. It uses zookeeper for consistent mapping across all
+ * nodes.
+ */
+public class GiraphClassResolver extends DefaultClassResolver {
+  /** Length of the ZK sequence number */
+  private static final int SEQUENCE_NUMBER_LENGTH = 10;
```
majakabiljo: Where is this number coming from?

giraph-core/src/main/java/org/apache/giraph/writable/kryo/GiraphClassResolver.java
```diff
@@ -0,0 +1,308 @@
...
+      throw new IllegalStateException(
+        "Failed to retrieve child nodes for " + KRYO_REGISTERED_CLASS_PATH, e);
+    } catch (InterruptedException e) {
+      throw new IllegalStateException(
+        "Interrupted while retrieving child nodes for " +
+                KRYO_REGISTERED_CLASS_PATH, e);
+    }
+
+    for (String name : registeredList) {
+      String className = name.substring(0,
```
majakabiljo: So when we create PERSISTENT_SEQUENTIAL node, zookeeper adds the sequential number to its file name? Maybe just add a comment stating that

giraph-core/src/main/java/org/apache/giraph/writable/kryo/GiraphClassResolver.java
```diff
@@ -0,0 +1,308 @@
...
+ * nodes.
+ */
+public class GiraphClassResolver extends DefaultClassResolver {
+  /** Length of the ZK sequence number */
+  private static final int SEQUENCE_NUMBER_LENGTH = 10;
+  /** Base ID to start for class name assignments.
+   * This number has to be high enough to not conflict with
+   * explicity registered class IDs.
+   * */
+  private static final int BASE_CLASS_ID = 10000;
```
majakabiljo: Is having this number high potentially increasing the serialized size of our data? Eg if it was low we could use a single byte to write class. Any way to check what is the highest number registered manually?

giraph-core/src/main/java/org/apache/giraph/writable/kryo/GiraphClassResolver.java
```diff
@@ -0,0 +1,308 @@
...
+    }
+
+    if (!ID_TO_CLASS_NAME.containsKey(id)) {
+      throw new IllegalStateException("ID " + id + " doesn't exist");
+    }
+    return ID_TO_CLASS_NAME.get(id);
+  }
+
+  @Override
+  public Registration register(Registration registration) {
```
majakabiljo, line 244: How does this method differ from implementation in DefaultClassResolver?

giraph-core/src/main/java/org/apache/giraph/writable/kryo/GiraphClassResolver.java
```diff
@@ -0,0 +1,308 @@
...
+    if (type == null) {
+      output.writeVarInt(Kryo.NULL, true);
+      return null;
+    }
+
+    Registration registration = kryo.getRegistration(type);
+    if (registration.getId() == NAME) {
+      throw new IllegalStateException("Invalid registration ID");
+    } else {
+      output.writeVarInt(registration.getId() + 2, true);
```
majakabiljo: What is the offset 2 for?

giraph-core/src/main/java/org/apache/giraph/writable/kryo/GiraphClassResolver.java
```diff
@@ -0,0 +1,308 @@
...
+  private static Map<Integer, String> ID_TO_CLASS_NAME = new HashMap();
+  /** Zookeeper */
+  private static ZooKeeperExt ZK;
+  /** Zookeeper path for automatic class registrations */
+  private static String KRYO_REGISTERED_CLASS_PATH;
+  /** Minimum class ID assigned by zookeeper sequencing */
+  private static int MIN_CLASS_ID = -1;
+
+  /** Memoized class id*/
+  private int memoizedClassId = -1;
```
majakabiljo: Is this to speed up the case when single class is used very frequently?

giraph-core/src/main/java/org/apache/giraph/writable/kryo/GiraphClassResolver.java
yukselakinci: The default implementation also handles the case when the class is non-explicitly registered, i.e. it is not assigned to an an integer. Since in this implementation all classes are assigned to integers agreed by all kryo instances, they all treated like explicitly assigned classes. This implementation throws an exception if it encounters a class that has not been assigned to an integer yet.

giraph-core/src/main/java/org/apache/giraph/writable/kryo/GiraphClassResolver.java
yukselakinci: Class ID's are incremented by 2 when writing, because 0 is used for null, and 1 is used for non-explicitly registered classes, though non-explicit class registration is an error for this implementation.

giraph-core/src/main/java/org/apache/giraph/writable/kryo/GiraphClassResolver.java
yukselakinci: Correct.

giraph-core/src/main/java/org/apache/giraph/writable/kryo/GiraphClassResolver.java
yukselakinci: The highest I observed was ~250. So 1000 would be more than enough.

giraph-core/src/main/java/org/apache/giraph/writable/kryo/GiraphClassResolver.java
yukselakinci: Done.

giraph-core/src/main/java/org/apache/giraph/writable/kryo/GiraphClassResolver.java
yukselakinci: It is coming from ZooKeeperExt, but was a private field. I changed it to public.

giraph-core/src/main/java/org/apache/giraph/writable/kryo/GiraphClassResolver.java
yukselakinci: Without static, we have to pass this information as a parameter. But there are so many places where the kryo objects are created, so it would not be pretty to delegate this information from BspService to all these places.

giraph-core/src/main/java/org/apache/giraph/writable/kryo/GiraphClassResolver.java
yukselakinci: Added the comments.

majakabiljo (approved):

giraph-core/src/main/java/org/apache/giraph/writable/kryo/GiraphClassResolver.java
dlogothetis: Can we add this as a comment here?

giraph-core/src/main/java/org/apache/giraph/writable/kryo/GiraphClassResolver.java
```diff
@@ -0,0 +1,333 @@
...
+import java.util.Map;
+import java.util.List;
+
+import static com.esotericsoftware.kryo.util.Util.getWrapperClass;
+
+/**
+ * This class resolver assigns unique classIds for every class that was not
+ * explicitly registered. It uses zookeeper for consistent mapping across all
+ * nodes.
+ */
```
dlogothetis: Can we add some more documentation the describes how this mechanism works?

giraph-core/src/main/java/org/apache/giraph/writable/kryo/GiraphClassResolver.java
```diff
@@ -0,0 +1,333 @@
...
+    }
+  }
+
+  /**
+   * Refreshes class-ID mapping from zookeeper.
+   * Not thread safe.
+   */
+  public static void refreshCache() {
+    try {
+      ZK.createOnceExt(KRYO_REGISTERED_CLASS_PATH,
```
dlogothetis: I think you can avoid calling this every time you refresh the cache and, instead, call it once during initialization (e.g. in setZookeeperInfo.

giraph-core/src/main/java/org/apache/giraph/writable/kryo/GiraphClassResolver.java
```diff
@@ -0,0 +1,333 @@
...
+                      false);
+    } catch (KeeperException e) {
+      throw new IllegalStateException(
+        "Failed to retrieve child nodes for " + KRYO_REGISTERED_CLASS_PATH, e);
+    } catch (InterruptedException e) {
+      throw new IllegalStateException(
+        "Interrupted while retrieving child nodes for " +
+                KRYO_REGISTERED_CLASS_PATH, e);
+    }
+
```
dlogothetis, line 170: Can we add some debug logging here that prints the returned list? With the usual pattern:

if (LOG.isDebugEnabled()) {
      LOG.debug("...");
}

giraph-core/src/main/java/org/apache/giraph/writable/kryo/GiraphClassResolver.java
```diff
@@ -0,0 +1,333 @@
...
+import static com.esotericsoftware.kryo.util.Util.getWrapperClass;
+
+/**
+ * This class resolver assigns unique classIds for every class that was not
+ * explicitly registered. It uses zookeeper for consistent mapping across all
+ * nodes.
+ */
+public class GiraphClassResolver extends DefaultClassResolver {
+  /** Base ID to start for class name assignments.
+   * This number has to be high enough to not conflict with
```
dlogothetis: Is a conflict something we can detect?

giraph-core/src/main/java/org/apache/giraph/writable/kryo/GiraphClassResolver.java
yukselakinci: Ok.

giraph-core/src/main/java/org/apache/giraph/writable/kryo/GiraphClassResolver.java
yukselakinci: Added more documentation.

yukselakinci: Added more documentation.

giraph-core/src/main/java/org/apache/giraph/writable/kryo/GiraphClassResolver.java
yukselakinci: We could if this information was not kept private in the base class. The application will certainly fail if there is a conflict.

giraph-core/src/main/java/org/apache/giraph/writable/kryo/GiraphClassResolver.java
yukselakinci: I have added a boolean flag to avoid this repeated call.

giraph-core/src/main/java/org/apache/giraph/writable/kryo/GiraphClassResolver.java
yukselakinci: Ok.

giraph-core/src/main/java/org/apache/giraph/writable/kryo/GiraphClassResolver.java
yukselakinci: Ok.

yukselakinci:
The custom class resolver allows kryo to always use the integer IDs for classes that are agreed by all nodes running the job. Default class resolver always writes the full class name of  the first encountered class type to the stream, and then it assigns an integer for subsequent instances. These changes make the serialization faster by eliminating the need to write the full name for the first encountered class instance. 

Tests:
1- Unit tests
2- Snapshot tests
3- Ran with two production pipelines
     Running time for mutual friends pipeline decreased from 41 to 35 minutes.

dlogothetis (approved):

PR #71, opened 2018-04-25
[PageRank] Make one-to-many messaging configurable

dlogothetis:
https://issues.apache.org/jira/browse/GIRAPH-1189

The PageRank implementation uses one-to-many messaging to reduce the amount of traffic sent. The corresponding message request (SendWorkerOneMessageToManyRequest)  creates ByteArrayVertexIdMessages instances before adding the messages to the Message Store. When the number of messages is big, the creation of this instances can fail. 

Adding an option here to enable/disable the use of one-to-many messaging.

Tests:
- Run with dataset that would cause PageRank to fail because of large number of messages.


majakabiljo:
Why do we have to create ByteArrayVertexIdMessages before adding to MessageStore, would it be possible to fix that instead and then we get benefit on all apps?

dlogothetis:
A more general solution at https://github.com/apache/giraph/pull/72.

PR #72, opened 2018-04-26
Improve memory usage of SendWorkerOneMessageToManyRequest

dlogothetis:
https://issues.apache.org/jira/browse/GIRAPH-1190

The current implementation takes incoming messages stored as ByteArrayOneMessageToManyIds and prepares them as a map from partition id to a ByteArrayVertexIdMessages, which holds the messages for the corresponding partition.  It then adds these to the message store.

However, it is possible that these intermediate lists of message get big before they are added to the message store. If they reach the capacity of the underlying buffers, the job fails. This can be avoided if we push these lists to the message store before the get big. This is mostly beneficial when we use a combiner in which case the message store keeps only one value per vertex.

Tests:
- Unit tests
- Compared performance on a large dataset and it is similar to the existing implementation.
- Verified that job that would otherwise fail, it succeeds.

### DIFF
--- a/giraph-core/src/test/java/org/apache/giraph/comm/RequestTest.java
+++ b/giraph-core/src/test/java/org/apache/giraph/comm/RequestTest.java
@@ -290,12 +290,8 @@ public class RequestTest {
     public int numVertices;
 
     @Parameterized.Parameters(name = "{index}: numVertices={0}")
-    public static Collection<Object[]> data() {
-      return Arrays.asList(new Object[][]{
-        {10},
-        {1000},
-        {1000000}
-      });
+    public static Object[] data() {
+      return new Object[] {10, 1000, 1000000};
     }
 
     @Test
--- a/giraph-core/src/test/java/org/apache/giraph/comm/RequestTest.java
+++ b/giraph-core/src/test/java/org/apache/giraph/comm/RequestTest.java
@@ -20,7 +20,6 @@ package org.apache.giraph.comm;
 
 import org.apache.giraph.comm.netty.NettyClient;
 import org.apache.giraph.comm.netty.NettyServer;
-import org.apache.giraph.comm.netty.handler.AckSignalFlag;
 import org.apache.giraph.comm.netty.handler.WorkerRequestServerHandler;
 import org.apache.giraph.comm.requests.SendPartitionMutationsRequest;
 import org.apache.giraph.comm.requests.SendVertexRequest;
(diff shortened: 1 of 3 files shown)


giraph-core/src/test/java/org/apache/giraph/comm/RequestTest.java
```diff
@@ -67,251 +71,280 @@
...
+
+  /**
+   * Tests that we run for different input parameters.
+   */
+  @RunWith(Parameterized.class)
+  public static class ParameterizedTest extends SharedSetup {
+
+    @Parameterized.Parameter(value = 0)
+    public int numVertices;
+
```
yukselakinci, line 291: Looks like if you only need a single parameter, you can just use:
@Parameters
public static Object[] data() {
    return new Object[] { "first test", "second test" };
}

https://github.com/junit-team/junit4/wiki/parameterized-tests

giraph-core/src/test/java/org/apache/giraph/comm/RequestTest.java
```diff
@@ -67,251 +71,280 @@
...
         Lists.<WorkerInfo>newArrayList(workerInfo));
+    }
   }
 
-  @Test
-  public void sendVertexPartition() {
-    // Data to send
-    int partitionId = 13;
-    Partition<IntWritable, IntWritable, IntWritable> partition =
+  public static class NonParameterizedTest extends SharedSetup {
```
yukselakinci, line 117: Any reason for not using Parameterized tests for this two test cases?

yukselakinci (approved):

yukselakinci:
Not sure why the commits show Graph-1185. Do you have to rebase?

giraph-core/src/test/java/org/apache/giraph/comm/RequestTest.java
dlogothetis: These other tests are a bit different with respect to what the parameter would be and the range of values. For now, I just wanted to make this particular test parameterized.

giraph-core/src/test/java/org/apache/giraph/comm/RequestTest.java
dlogothetis: I'll simplify this then, thanks!

dlogothetis:
I accidentally created a commit with the same message as the previous commit. When this lands, it'll get a new commit message anyway.

yukselakinci (changes requested):
Needs better unit tests.

PR #76, opened 2018-05-30
Register serializers for ImmutableBiMap in HadoopKryo

dlogothetis:
https://issues.apache.org/jira/browse/GIRAPH-1195

When using guava, RegularImmutableBiMap is may be used underneath the covers as an immutable map implementation. The default MapSerializer does not work for this implementation as it does not support the put() operation. Instead, we can register this class to use the ImmutableMapSerializer just like with RegularImmutableMap.

This PR updates Kryo and the kryo-serializers dependencies to the latest version. kryo-serializers has some utility methods for registering map classes.

Tests:
- Unit tests


giraph-core/src/main/java/org/apache/giraph/writable/kryo/HadoopKryo.java
```diff
@@ -322,7 +323,8 @@ private static HadoopKryo createKryo(boolean trackReferences,
     if (minor >= '8') {
       try {
         kryo.register(Class.forName("java.lang.invoke.SerializedLambda"));
-        kryo.register(Class.forName("com.esotericsoftware.kryo.Kryo$Closure"),
```
yukselakinci, line 325: Is this replaced with ClouseSerializer$Closure in new version?

giraph-core/src/main/java/org/apache/giraph/writable/kryo/serializers/ImmutableBiMapSerializerUtils.java
```diff
@@ -0,0 +1,60 @@
...
+   */
+  public static void registerSerializers(Kryo kryo) {
+    ImmutableMapSerializer serializer = new ImmutableMapSerializer();
+    kryo.register(ImmutableBiMap.class, serializer);
+    kryo.register(ImmutableBiMap.of().getClass(), serializer);
+    Object o1 = new Object();
+    Object o2 = new Object();
+    Object o3 = new Object();
+    Object o4 = new Object();
+    Object o5 = new Object();
```
yukselakinci, line 50: Are the following implementations all used?

giraph-core/src/test/java/org/apache/giraph/writable/kryo/KryoWritableWrapperTest.java
```diff
@@ -170,6 +173,41 @@ public void testImmutableList() throws IOException {
     }
   }
 
+  @Test
+  public void testImmutableMapSerialization() throws IOException {
+    Map original = ImmutableMap.of("x", "y", "y", "z");
+    Map copy = kryoSerDeser(original);
+    assertEquals(original, copy);
```
yukselakinci, line 180: Does this assert individually compare all the properties?

giraph-core/src/main/java/org/apache/giraph/writable/kryo/serializers/ImmutableBiMapSerializerUtils.java
```diff
@@ -0,0 +1,60 @@
...
+  private ImmutableBiMapSerializerUtils() {
+  }
+
+  /**
+   * Registers different {@link ImmutableBiMap} implementations with Kryo.
+   * @param kryo {@link Kryo} instance to register class with.
+   */
+  public static void registerSerializers(Kryo kryo) {
+    ImmutableMapSerializer serializer = new ImmutableMapSerializer();
+    kryo.register(ImmutableBiMap.class, serializer);
```
yukselakinci, line 44: Do you know what is the reason behind ImmutableBiMap using the same serializer as ImmutableMap?

giraph-core/src/main/java/org/apache/giraph/writable/kryo/HadoopKryo.java
dlogothetis: That's right.

giraph-core/src/main/java/org/apache/giraph/writable/kryo/serializers/ImmutableBiMapSerializerUtils.java
dlogothetis: They are not used currently, but if somebody ever uses them, this should cover these cases.

giraph-core/src/test/java/org/apache/giraph/writable/kryo/KryoWritableWrapperTest.java
dlogothetis: Yes, this compares key-value pair equality.

giraph-core/src/main/java/org/apache/giraph/writable/kryo/serializers/ImmutableBiMapSerializerUtils.java
dlogothetis: An BiMap is actually a Map with some extra constraints on the values of the map entries: https://google.github.io/guava/releases/19.0/api/docs/com/google/common/collect/BiMap.html. So the MapSerializer should work for this one too.

yukselakinci (approved):

PR #78, opened 2018-07-30
Fixes Incorrect Logging Statement on #255 in GiraphYarnClient

Swapnil-Gandhi:
Summary :

In-case, Yarn-Heap-Size passed in Giraph Configuration exceeds max available capacity, we downgrade its value to max capacity (Lines #253-#257). 

Currently, the logger message associated with this conditional statement incorrectly states "minimum" instead of "maximum", which is misleading. This change fixes it.


Swapnil-Gandhi:
JIRA Issue : https://issues.apache.org/jira/browse/GIRAPH-1199

dlogothetis (approved):

PR #79, opened 2018-08-23
Prefers OutputStream instead of FileOutputStream to improve performance.

reudismam:


dlogothetis:
Can you comment on how this improves performance?

reudismam:
FileInputStream and FileOutputStream override the finalize method. As a result, their objects are only cleaned when the garbage collector performs a sweep. Since Java 7, programmers can use Files.newInputStream and Files.newOutputStream instead of FileInputStream and FileOutputStream to improve performance. finalize method is deprecated in Java 9+.

dlogothetis (approved):

dlogothetis (commented):
@reudismam, how did you test this change?

PR #81, opened 2018-08-23
Use the class Path to represent a path instead of a String.

reudismam:


dlogothetis:
@reudismam, can you create a JIRA issue on https://issues.apache.org/jira/projects/GIRAPH/ for this and your other pull requests? Also, could you add in the description of the PRs how you tested them? Thanks!

dlogothetis:
Can also describe why this change is necessary?

reudismam:
Strings can be used to represent a file system path even though some classes are specifically designed for this task. For instance, java.nio.Path. It is useful to change the type of the variable to Path, since strings can be combined in an undisciplined way, which can lead to invalid paths. Second, different operating systems use different file separators, which can cause bugs.

PR #83, opened 2018-09-15
GIRAPH-1200: Add counters for network request resends

majakabiljo:
Summary: Expose statistics around network requests which we had to resend.

Test Plan: Ran a normal job, these counters were 0. Ran a job with very low request timeouts and verified these counters got populated.

### DIFF
--- a/giraph-core/src/main/java/org/apache/giraph/comm/netty/NettyClient.java
+++ b/giraph-core/src/main/java/org/apache/giraph/comm/netty/NettyClient.java
@@ -38,6 +38,7 @@ import org.apache.giraph.comm.requests.WritableRequest;
 import org.apache.giraph.conf.BooleanConfOption;
 import org.apache.giraph.conf.GiraphConstants;
 import org.apache.giraph.conf.ImmutableClassesGiraphConfiguration;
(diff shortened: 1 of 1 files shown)


giraph-core/src/main/java/org/apache/giraph/comm/netty/NettyClient.java
```diff
@@ -966,22 +991,23 @@ public boolean apply(RequestInfo requestInfo) {
...
   private void resendRequestsWhenNeeded(
-      Predicate<RequestInfo> shouldResendRequestPredicate) {
+      Predicate<RequestInfo> shouldResendRequestPredicate,
+      GiraphHadoopCounter counter) {
     // Check if there are open requests which have been sent a long time ago,
     // and if so, resend them.
     List<ClientRequestId> addedRequestIds = Lists.newArrayList();
     List<RequestInfo> addedRequestInfos = Lists.newArrayList();
-    // Check all the requests for problems
+    // Check all the requests for problemsAggregate vertices
```
dlogothetis: This comment doesn't look right

dlogothetis (approved):

PR #84, opened 2018-09-18
Fix BspServiceMaster bug

dlogothetis:
https://issues.apache.org/jira/browse/GIRAPH-1201

Occasionally, BspServiceMaster fails to detect that a worker finished a superstep (e.g. input superstep, or a compute supestep). As a result, the job may get stuck as the master keeps waiting for all workers to finish.

While this is hard to reproduce, this fixes a potential cause for this.

Tests
- Unit tests
- Internal snapshot tests
- Ran large job

### DIFF
--- a/giraph-core/src/main/java/org/apache/giraph/master/BspServiceMaster.java
+++ b/giraph-core/src/main/java/org/apache/giraph/master/BspServiceMaster.java
@@ -1379,9 +1379,15 @@ public class BspServiceMaster<I extends WritableComparable,
 
       // Wait for a signal or timeout
       boolean eventTriggered = event.waitMsecs(eventLoopTimeout);
+
+      // If the event was triggered, we reset it. In the next loop run, we will
+      // read ZK to get the new hosts.
+      if (eventTriggered) {
+        event.reset();
+      }
+
       long elapsedTimeSinceRegularRunMsec = System.currentTimeMillis() -
           lastRegularRunTimeMsec;
-      event.reset();
       getContext().progress();
 
       if (eventTriggered ||


giraph-core/src/main/java/org/apache/giraph/master/BspServiceMaster.java
```diff
@@ -1379,9 +1379,15 @@ private boolean barrierOnWorkerList(String finishedWorkerPath,
...
+
+      // If the event was triggered, we reset it. In the next loop run, we will
+      // read ZK to get the new hosts.
+      if (eventTriggered) {
+        event.reset();
+      }
+
       long elapsedTimeSinceRegularRunMsec = System.currentTimeMillis() -
           lastRegularRunTimeMsec;
-      event.reset();
```
dlogothetis, line 1384: It's possible that after `event.waitMsecs` exits (due to timeout) and before `event.reset()` get called, the event get signaled. In this case, in the next loop `event.waitMsec` will timeout again and `logInfoOnlyRun` will continue to be false.

majakabiljo: Making sure I understand what's happening, does reodering of commands here makes no difference and the fix is about only clearing event if it triggered?

dlogothetis: Are you talking about re-ordering the `long elapsedTimeSinceRegularRunMsec = ` command and the `event.reset()` command? That shouldn't matter, I just moved the `if(eventTriggered)` because is seemed a bit clearer to put it close to the code that outputs `eventTriggered`.  The fix is the actual `if(eventTriggered)`.

majakabiljo (approved):

PR #88, opened 2018-10-15
GIRAPH-1205: Separate Giraph counters for different causes of network…

atanu1991:
In https://issues.apache.org/jira/browse/GIRAPH-1200 we added counters for network requests resends, but "network resends for timeout" actually cover various reasons for why request can get resent. We suspect that resending requests after maxRequestMilliseconds timeout is what is always fatal, so separating this counter from the other causes will help us verify that.

### DIFF
--- a/giraph-core/src/main/java/org/apache/giraph/comm/netty/NettyClient.java
+++ b/giraph-core/src/main/java/org/apache/giraph/comm/netty/NettyClient.java
@@ -141,6 +141,9 @@ public class NettyClient {
   /** How many network requests were resent because channel failed */
   public static final String NETWORK_REQUESTS_RESENT_FOR_CHANNEL_FAILURE_NAME =
       "Network requests resent for channel failure";
+  /** How many network requests were resent because connection failed */
+  public static final String NETWORK_REQUESTS_RESENT_FOR_CONNECTION_FAILURE_NAME =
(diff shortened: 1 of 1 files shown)


atanu1991:
Test plan:
Tested internally by running hello page rank 
drop=true ./hellopagerank.sh -DhelloPageRank.iterations=5000 -Dgiraph.maxRequestMilliseconds=1 -Dgiraph.waitingRequestMsecs=1

The new counters are seen
screen shot 2018-10-15 at 10 38 57 am

giraph-core/src/main/java/org/apache/giraph/comm/netty/NettyClient.java
```diff
@@ -141,6 +141,9 @@
   /** How many network requests were resent because channel failed */
   public static final String NETWORK_REQUESTS_RESENT_FOR_CHANNEL_FAILURE_NAME =
       "Network requests resent for channel failure";
+  /** How many network requests were resent because connection failed */
+  public static final String NETWORK_REQUESTS_RESENT_FOR_CONNECTION_FAILURE =
```
majakabiljo: Nit: NETWORK_REQUESTS_RESENT_FOR_CONNECTION_FAILURE_NAME

majakabiljo (approved):

PR #93, opened 2018-11-07
GIRAPH-1211: Make retrying to send network requests after timeout optional

majakabiljo:
Summary: Using counters added in GIRAPH-1205 we were able to confirm that resending network requests after timeout almost never succeeds, so add an option to fail early instead of keep trying to resend these network requests indefinitely.

Test Plan: Made response not being sent for a specific request id, verified that with keeping default value for this new option we kept trying to resend the request, and when making it false the job failed when request timeout was noticed.

### DIFF
--- a/giraph-core/src/main/java/org/apache/giraph/comm/netty/NettyClient.java
+++ b/giraph-core/src/main/java/org/apache/giraph/comm/netty/NettyClient.java
@@ -94,6 +94,7 @@ import static org.apache.giraph.conf.GiraphConstants.NETTY_CLIENT_EXECUTION_AFTE
 import static org.apache.giraph.conf.GiraphConstants.NETTY_CLIENT_EXECUTION_THREADS;
 import static org.apache.giraph.conf.GiraphConstants.NETTY_CLIENT_USE_EXECUTION_HANDLER;
 import static org.apache.giraph.conf.GiraphConstants.NETTY_MAX_CONNECTION_FAILURES;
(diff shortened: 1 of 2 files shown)


giraph-core/src/main/java/org/apache/giraph/conf/GiraphConstants.java
```diff
@@ -690,6 +690,12 @@
       new IntConfOption("giraph.maxRequestMilliseconds", MINUTES.toMillis(10),
           "Milliseconds for a request to complete (or else resend)");
 
+  /** Milliseconds for a request to complete (or else resend) */
```
dlogothetis: Comment seems inaccurate.

giraph-core/src/main/java/org/apache/giraph/comm/netty/NettyClient.java
```diff
@@ -1006,18 +1013,21 @@ public boolean apply(RequestInfo requestInfo) {
...
+    }, networkRequestsResentForConnectionFailure, true);
   }
 
   /**
    * Resend requests which satisfy predicate
    *  @param shouldResendRequestPredicate Predicate to use to check whether
    *                                     request should be resent
    * @param counter Counter to increment for every resent network request
+   * @param resendProblematicRequest Whether to resend problematic request or
+   *                                fail th job if such request is found
```
dlogothetis:
```suggestion
   *                                fail the job if such request is found
```

dlogothetis (approved):

PR #96, opened 2018-11-29
GIRAPH-1213: Fix issues with network requests retries and add more logging

majakabiljo:
Fixing two bugs:
- When channel fails, we are currently retrying all requests towards the destination machine from the channel, instead of just ones which are happening on the concrete channel.
- In practice, we've noticed BlockingOperationException can get thrown when we wait to connect on channel in which case we silently don't send the request we are trying to send, so catching this exception and retrying instead.
Also added logging of channel ids to be able to debug issues related to network requests not delivering easier.

### DIFF
--- a/giraph-core/src/main/java/org/apache/giraph/comm/netty/NettyClient.java
+++ b/giraph-core/src/main/java/org/apache/giraph/comm/netty/NettyClient.java
@@ -82,6 +82,7 @@ import io.netty.handler.codec.FixedLengthFrameDecoder;
 import io.netty.handler.codec.LengthFieldBasedFrameDecoder;
 import io.netty.util.AttributeKey;
 /*end[HADOOP_NON_SECURE]*/
+import io.netty.util.concurrent.BlockingOperationException;
 import io.netty.util.concurrent.DefaultEventExecutorGroup;
 import io.netty.util.concurrent.EventExecutorGroup;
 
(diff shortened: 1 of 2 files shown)


giraph-core/src/main/java/org/apache/giraph/comm/netty/NettyClient.java
```diff
@@ -1147,8 +1158,11 @@ private void checkRequestsAfterChannelFailure(final Channel channel) {
     resendRequestsWhenNeeded(new Predicate<RequestInfo>() {
       @Override
       public boolean apply(RequestInfo requestInfo) {
-        return requestInfo.getDestinationAddress().equals(
-            channel.remoteAddress());
+        if (requestInfo.getWriteFuture() == null ||
```
dlogothetis, line 1161: When is this condition true?

dlogothetis:
Can you also add a couple of comments about how this was tested?

giraph-core/src/main/java/org/apache/giraph/comm/netty/NettyClient.java
majakabiljo: It can happen if the request wasn't sent out yet, not sure if there is some other scenario.

majakabiljo:
I used a pipeline which runs 100 jobs and was always getting at least a few jobs stuck with open network requests. Running it with more logging helped identify these two issues, and after the change it was 100% successful.

dlogothetis (approved):

PR #101, opened 2019-05-08
Make number of retries in RetryableJobProgressTrackerClient configurable

dlogothetis:
Currently, the number of retries is fixed to 1. This makes the number of retries and the time to wait between retries configurable.

https://issues.apache.org/jira/projects/GIRAPH/issues/GIRAPH-1218

Tests
- mvn -Phadoop_2 clean install
- Ran a number of jobs

Looking at the logs, I verified that some failed methods get retried with a wait time between calls.

### DIFF
--- a/giraph-core/src/main/java/org/apache/giraph/graph/RetryableJobProgressTrackerClient.java
+++ b/giraph-core/src/main/java/org/apache/giraph/graph/RetryableJobProgressTrackerClient.java
@@ -69,7 +69,7 @@ public class RetryableJobProgressTrackerClient
   /** Job progress tracker */
   private JobProgressTracker jobProgressTracker;
   /** Cached value for number of retries */
-  private int numRetries = 1;
+  private int numRetries;
   /** Cached value for wait time between retries */
   private int retryWaitMsec;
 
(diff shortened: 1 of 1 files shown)


giraph-core/src/main/java/org/apache/giraph/graph/RetryableJobProgressTrackerClient.java
```diff
@@ -57,6 +68,10 @@
   private ThriftClientManager clientManager;
   /** Job progress tracker */
   private JobProgressTracker jobProgressTracker;
+  /** Cached value for number of retries */
+  private int numRetries = 1;
+  /** Cached value for wait time between retries */
+  private int retryWaitMsec;
```
atanu1991, line 74: Do we need to initialize this to 1000?

atanu1991 (approved):

giraph-core/src/main/java/org/apache/giraph/graph/RetryableJobProgressTrackerClient.java
dlogothetis: We don't need to. It will get initialized in the constructor. For consistency in the style, I will remove the initialization of numRertries too.

PR #107, opened 2019-09-06
Configure jmap path

aanchal204:
GIRAPH-1223

### DIFF
--- a/giraph-core/src/main/java/org/apache/giraph/utils/JMap.java
+++ b/giraph-core/src/main/java/org/apache/giraph/utils/JMap.java
@@ -30,8 +30,6 @@ import java.util.Date;
  * Helper to run jmap and print the output
  */
 public class JMap {
-  /** The command to run */
-  public static final String CMD = "jmap ";
   /** Arguments to pass in to command */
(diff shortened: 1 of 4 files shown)


giraph-core/src/main/java/org/apache/giraph/utils/JMap.java
```diff
@@ -89,13 +94,16 @@ public static void heapHistogramDump(int numLines, PrintStream printStream) {
...
    */
   private static void heapHistogramDump(int numLines,
                                         boolean liveObjectsOnly,
-                                        PrintStream printStream) {
+                                        PrintStream printStream,
+                                        String jmapPath) {
     try {
       String args = liveObjectsOnly ? LIVE_HISTO_OPTION : ARGS;
-      Process p = Runtime.getRuntime().exec(CMD + args + getProcessId());
+      String cmd = jmapPath.isEmpty() ? CMD : (jmapPath + CMD);
```
majakabiljo: Nit: maybe make the conf option JMAP_LOCATION with default "jmap"

majakabiljo (approved):
lg

PR #110, opened 2019-10-26
Counter Mechanism

aanchal204:
Counter mechanism
Adds the mechanism to generate counters like gc-related metrics, file system metrics and job-specific counters. 
The counters need to be defined in CustomCounters, when storing in the job context.
The workers write these counters to the zookeeper after each superstep, and the master aggregates them from all the workers, and sends these aggregated counters to the job client using a thrift client.
 
In case of a failure, the master would have the aggregated counters until the previous superstep, and these partial counters will be sent using the thrift client.

### DIFF
--- a/giraph-core/src/main/java/org/apache/giraph/bsp/BspService.java
+++ b/giraph-core/src/main/java/org/apache/giraph/bsp/BspService.java
@@ -92,6 +92,10 @@ public abstract class BspService<I extends WritableComparable,
   public static final String MASTER_ELECTION_DIR = "/_masterElectionDir";
   /** Superstep scope */
   public static final String SUPERSTEP_DIR = "/_superstepDir";
+  /** Counter sub directory */
+  public static final String COUNTERS_DIR = "/_counters";
+  /** Metrics sub directory */
+  public static final String METRICS_DIR = "/_metrics";
   /** Healthy workers register here. */
   public static final String WORKER_HEALTHY_DIR = "/_workerHealthyDir";
   /** Unhealthy workers register here. */
@@ -248,8 +252,10 @@ public abstract class BspService<I extends WritableComparable,
     this.graphPartitionerFactory = conf.createGraphPartitioner();
 
     basePath = ZooKeeperManager.getBasePath(conf) + BASE_DIR + "/" + jobId;
-    getContext().getCounter(GiraphConstants.ZOOKEEPER_BASE_PATH_COUNTER_GROUP,
-        basePath);
+    if (LOG.isInfoEnabled()) {
+      LOG.info(String.format("%s: %s",
+              GiraphConstants.ZOOKEEPER_BASE_PATH_COUNTER_GROUP, basePath));
+    }
     masterJobStatePath = basePath + MASTER_JOB_STATE_NODE;
     inputSplitsWorkerDonePath = basePath + INPUT_SPLITS_WORKER_DONE_DIR;
     inputSplitsAllDonePath = basePath + INPUT_SPLITS_ALL_DONE_NODE;
@@ -425,11 +431,20 @@ public abstract class BspService<I extends WritableComparable,
    *
    * @param attempt application attempt number
    * @param superstep superstep to use
+   * @param readWriteCounters Boolean to denote whether we are
+   *                          reading/writing counters
    * @return directory path based on the a superstep
    */
-  public final String getWorkerFinishedPath(long attempt, long superstep) {
-    return applicationAttemptsPath + "/" + attempt +
+  public final String getWorkerFinishedPath(long attempt, long superstep,
+                                            boolean readWriteCounters) {
+    String finishedDir = applicationAttemptsPath + "/" + attempt +
         SUPERSTEP_DIR + "/" + superstep + WORKER_FINISHED_DIR;
+    if (readWriteCounters) {
+      finishedDir += COUNTERS_DIR;
+    } else {
+      finishedDir += METRICS_DIR;
+    }
+    return finishedDir;
   }
 
   /**
--- a/giraph-core/src/main/java/org/apache/giraph/bsp/BspService.java
+++ b/giraph-core/src/main/java/org/apache/giraph/bsp/BspService.java
@@ -944,7 +944,8 @@ public abstract class BspService<I extends WritableComparable,
       eventProcessed = true;
     } else if (event.getPath().endsWith(COUNTERS_DIR) &&
             event.getType() == EventType.NodeChildrenChanged) {
-      writtenCountersToZK.signal();
+      LOG.info("process: writtenCountersToZK signaled");
+      getWrittenCountersToZKEvent().signal();
       eventProcessed = true;
     }
 
--- a/giraph-core/src/main/java/org/apache/giraph/bsp/BspService.java
+++ b/giraph-core/src/main/java/org/apache/giraph/bsp/BspService.java
@@ -182,6 +182,9 @@ public abstract class BspService<I extends WritableComparable,
   private final BspEvent masterElectionChildrenChanged;
   /** Cleaned up directory children changed*/
   private final BspEvent cleanedUpChildrenChanged;
+  /** Event to synchronize when workers have written their counters to the
+   * zookeeper*/
+  private final BspEvent writtenCountersToZK;
   /** Registered list of BspEvents */
   private final List<BspEvent> registeredBspEvents =
       new ArrayList<BspEvent>();
@@ -227,6 +230,7 @@ public abstract class BspService<I extends WritableComparable,
     this.superstepFinished = new PredicateLock(context);
     this.masterElectionChildrenChanged = new PredicateLock(context);
     this.cleanedUpChildrenChanged = new PredicateLock(context);
+    this.writtenCountersToZK = new PredicateLock(context);
 
     registerBspEvent(connectedEvent);
     registerBspEvent(workerHealthRegistrationChanged);
@@ -236,6 +240,7 @@ public abstract class BspService<I extends WritableComparable,
     registerBspEvent(superstepFinished);
     registerBspEvent(masterElectionChildrenChanged);
     registerBspEvent(cleanedUpChildrenChanged);
+    registerBspEvent(writtenCountersToZK);
 
     this.context = context;
     this.graphTaskManager = graphTaskManager;
@@ -427,24 +432,31 @@ public abstract class BspService<I extends WritableComparable,
 
   /**
    * Generate the worker "finished" directory path for a
-   * superstep
+   * superstep, for storing the superstep-related metrics
    *
    * @param attempt application attempt number
    * @param superstep superstep to use
-   * @param readWriteCounters Boolean to denote whether we are
-   *                          reading/writing counters
    * @return directory path based on the a superstep
    */
-  public final String getWorkerFinishedPath(long attempt, long superstep,
-                                            boolean readWriteCounters) {
-    String finishedDir = applicationAttemptsPath + "/" + attempt +
-        SUPERSTEP_DIR + "/" + superstep + WORKER_FINISHED_DIR;
-    if (readWriteCounters) {
-      finishedDir += COUNTERS_DIR;
-    } else {
-      finishedDir += METRICS_DIR;
-    }
-    return finishedDir;
+  public final String getWorkerMetricsFinishedPath(
+          long attempt, long superstep) {
+    return applicationAttemptsPath + "/" + attempt +
+            SUPERSTEP_DIR + "/" + superstep + WORKER_FINISHED_DIR + METRICS_DIR;
+  }
+
+  /**
+   * Generate the worker "finished" directory path for a
+   * superstep, for storing the superstep-related counters
+   *
+   * @param attempt application attempt number
+   * @param superstep superstep to use
+   * @return directory path based on the a superstep
+   */
+  public final String getWorkerCountersFinishedPath(
+          long attempt, long superstep) {
+    return applicationAttemptsPath + "/" + attempt +
+            SUPERSTEP_DIR + "/" + superstep +
+            WORKER_FINISHED_DIR + COUNTERS_DIR;
   }
 
   /**
@@ -598,6 +610,10 @@ public abstract class BspService<I extends WritableComparable,
 
   public final BspEvent getCleanedUpChildrenChangedEvent() {
     return cleanedUpChildrenChanged;
+  }
+
+  public final BspEvent getWrittenCountersToZKEvent() {
+    return writtenCountersToZK;
   }
 
   /**
@@ -925,6 +941,10 @@ public abstract class BspService<I extends WritableComparable,
         LOG.info("process: cleanedUpChildrenChanged signaled");
       }
       cleanedUpChildrenChanged.signal();
+      eventProcessed = true;
+    } else if (event.getPath().endsWith(COUNTERS_DIR) &&
+            event.getType() == EventType.NodeChildrenChanged) {
+      writtenCountersToZK.signal();
       eventProcessed = true;
     }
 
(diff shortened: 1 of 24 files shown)


giraph-core/src/main/java/org/apache/giraph/counters/CustomCounters.java
```diff
@@ -0,0 +1,112 @@
...
+   *
+   * @return Map of unique counter names
+   */
+  public static Set<CustomCounter> getCustomCounters() {
+    return COUNTER_NAMES;
+  }
+
+  /**
+   * Merge the incoming counter values with the existing counters
+   *
```
dlogothetis, line 87: Let's add a comment about when this is expected to get called.

giraph-core/src/main/java/org/apache/giraph/worker/BspServiceWorker.java
```diff
@@ -1236,6 +1233,57 @@ public void cleanup(FinishedSuperstepStats finishedSuperstepStats)
...
           "to master about cleanup", e);
     }
+  }
+
+  /**
+   * Method to send the counter values to the master using the zookeeper
+   * This is called after finishing all the supersteps
+   *
+   */
+  public void sendCountersToMaster() {
```
dlogothetis: nit: since this is really storing counters in ZK, and not sending to master directly, let's change the name to reflect this.

giraph-core/src/main/java/org/apache/giraph/counters/CustomCounters.java
```diff
@@ -0,0 +1,112 @@
...
+ */
+public class CustomCounters {
+
+
+
+  /** Unique counter groups and names populated during execution of the job */
+  private static Set<CustomCounter> COUNTER_NAMES = new HashSet<>();
+
+  /** Aggregated counter values updated by master */
+  private final List<CustomCounter> counterList;
```
dlogothetis, line 49: Since the basic operation you're doing on the list is search and update the value, maybe make this a Set for faster lookups?

giraph-core/src/main/java/org/apache/giraph/counters/GiraphStats.java
```diff
@@ -151,6 +153,23 @@ public static GiraphStats getInstance() {
...
   }
 
+
+  /**
+   * Get a map of counter names, and values
+   * To be used by the master to send to the job client
+   *
+   * @return Map of counter names and values
+   */
+  public List<CustomCounter> getCounterMap() {
```
dlogothetis: nit: rename to `getCounterList()`?

giraph-core/src/main/java/org/apache/giraph/counters/GiraphTimers.java
```diff
@@ -180,4 +182,27 @@ public GiraphHadoopCounter getInitializeMs() {
     return Iterators.concat(jobCounters().iterator(),
         superstepCounters().values().iterator());
   }
+
+  /**
+   * Get a map of counter names and values
+   *
+   * @return Map of counter names and values
+   */
+  public List<CustomCounter> getCounterMap() {
```
dlogothetis: nit: rename to getCounterList()?

giraph-core/src/main/java/org/apache/giraph/counters/CustomCounters.java
```diff
@@ -0,0 +1,112 @@
...
+
+package org.apache.giraph.counters;
+
+import java.util.ArrayList;
+import java.util.HashSet;
+import java.util.List;
+import java.util.Set;
+
+/**
+ * Custom counters created by the client or app
```
dlogothetis, line 27: Let's add a few comments here about how we expect that whenever we create counters, that they need to be added/registered here as well. Also a few comments about why we need this new class.

giraph-core/src/main/java/org/apache/giraph/counters/CustomCounters.java
aanchal204: I cannot use a set because I need to be able to get the element, update its contents and insert it back into the set. But, hash set does not support a method to fetch an element from the set.

aanchal204 (commented):
Addressed comments and created a map of counters (wherever possible) to store the counter names, as opposed to storing them in custom counters

giraph-core/src/main/java/org/apache/giraph/bsp/BspService.java
```diff
@@ -425,11 +431,20 @@ public final String getWorkerWroteCheckpointPath(long attempt,
    *
    * @param attempt application attempt number
    * @param superstep superstep to use
+   * @param readWriteCounters Boolean to denote whether we are
+   *                          reading/writing counters
    * @return directory path based on the a superstep
    */
-  public final String getWorkerFinishedPath(long attempt, long superstep) {
-    return applicationAttemptsPath + "/" + attempt +
+  public final String getWorkerFinishedPath(long attempt, long superstep,
```
dlogothetis: Let's make a separate method that just returns the counters path. It seems clearer to separate these two.

giraph-core/src/main/java/org/apache/giraph/bsp/CentralizedServiceWorker.java
```diff
@@ -256,4 +256,19 @@ void cleanup(FinishedSuperstepStats finishedSuperstepStats)
    */
   void addressesAndPartitionsReceived(
       AddressesAndPartitionsWritable addressesAndPartitions);
+
+  /**
+   * Send the counter values to the master after every superstep
+   * and also after all supersteps are done. This is called before closing
```
dlogothetis: These comments don't describe exactly what the method does. We're not really sending to the master.

giraph-core/src/main/java/org/apache/giraph/counters/CustomCounter.java
```diff
@@ -0,0 +1,188 @@
...
+  private String counterName;
+
+  /** Type of aggregation for counter */
+  private AGGREGATION aggregation;
+
+  /** Counter value */
+  private long value;
+
+  /** To store the type of aggregation to be done on the counters */
+  public enum AGGREGATION {
```
dlogothetis: nit: i think all caps is not a common practice, maybe name just `Aggregation`?

giraph-core/src/main/java/org/apache/giraph/worker/BspServiceWorker.java
```diff
@@ -1236,6 +1235,76 @@ public void cleanup(FinishedSuperstepStats finishedSuperstepStats)
...
+        jsonCounters.put(Base64.encodeBytes(
+                WritableUtils.writeToByteArray(customCounter)));
+      }
+    }
+    long superStep = getSuperstep() + (allSuperstepsDone ? 1 : 0);
+    String finishedWorkerPath =
+            getWorkerFinishedPath(
+                    getApplicationAttempt(), superStep, true) +
+                    "/" + workerInfo.getHostnameId();
+    LOG.info("worker finished path: " + finishedWorkerPath);
```
dlogothetis: Related to the commend about having a separate method for getting the path for the counters, let's change the log line too, to reflect we're writing counters.

giraph-core/src/main/java/org/apache/giraph/master/BspServiceMaster.java
```diff
@@ -1794,6 +1804,154 @@ private void doMasterCompute() {
...
+   * The aggregated counters are stored in a thrift struct
+   */
+  private void aggregateCountersFromWorkersAndMaster() {
+    CustomCounters customCounters = new CustomCounters();
+    // Get the stats from the all the worker selected nodes
+    String workerFinishedPath = getWorkerFinishedPath(getApplicationAttempt(),
+            getSuperstep(), true);
+    List<String> workerFinishedPathList = null;
+    // Subtract 1 for the master
+    // TODO - what happens when there is only 1 worker?
```
dlogothetis: Please fix comment to reflect how you treat the case of 1 worker and why.

giraph-core/src/main/java/org/apache/giraph/master/BspServiceMaster.java
```diff
@@ -1794,6 +1804,154 @@ private void doMasterCompute() {
...
+        counters.add(customCounter);
+      }
+    }
+  }
+
+  /**
+   * Receive the counters from the workers, and aggregate them with the
+   * master counters.
+   * The aggregated counters are stored in a thrift struct
+   */
```
dlogothetis, line 1845: Let's add some comment about when this method is expected  to be called.

giraph-core/src/main/java/org/apache/giraph/master/BspServiceMaster.java
```diff
@@ -1794,6 +1804,154 @@ private void doMasterCompute() {
...
+    List<String> workerFinishedPathList = null;
+    // Subtract 1 for the master
+    // TODO - what happens when there is only 1 worker?
+    int numWorkers = BspInputFormat.getMaxTasks(getConfiguration()) - 1;
+    if (numWorkers == 0) {
+      numWorkers += 1;
+    }
+    // Get the counter values from the zookeeper, written by the workers
+    // We keep retrying until all the workers have written
+    // TODO- wait for definite time only
```
dlogothetis: There are utility methods in the giraph code that can loop with a timeout.

giraph-core/src/main/java/org/apache/giraph/master/BspServiceMaster.java
```diff
@@ -1794,6 +1804,154 @@ private void doMasterCompute() {
...
+                "aggregateWorkerStats: KeeperException", e);
+      } catch (InterruptedException e) {
+        throw new IllegalStateException(
+                "aggregateWorkerStats: InterruptedException", e);
+      } catch (IOException e) {
+        throw new IllegalStateException("aggregateWorkerStats: IOException", e);
+      }
+    }
+    // Add master counters too
+    // TODO: ensure counters related to master only also appear with w=1
```
dlogothetis: remove this comment?

giraph-core/src/main/java/org/apache/giraph/master/BspServiceMaster.java
```diff
@@ -1794,6 +1804,154 @@ private void doMasterCompute() {
...
+      populateCountersFromContext(context, inputSplitCounter, masterCounters);
+      customCounters.mergeCounters(masterCounters);
+    }
+    // Add GiraphStats
+    List<CustomCounter> allCounters = new ArrayList<>();
+    allCounters.addAll(GiraphStats.getInstance().getCounterList());
+    // Custom counters
+    allCounters.addAll(customCounters.getCounterList());
+    // Store in Thrift Struct
+    GiraphCountersThriftStruct.get().setCounters(allCounters);
```
dlogothetis: Not sure why there has to be a static instance of this. It seems that whenever you need send the counters, you can create a local instance of GiraphCountersThriftStruct and send it.

giraph-core/src/main/java/org/apache/giraph/worker/BspServiceWorker.java
```diff
@@ -797,6 +795,7 @@ public FinishedSuperstepStats finishSuperstep(
     }
     writeFinshedSuperstepInfoToZK(partitionStatsList,
       workerSentMessages, workerSentMessageBytes);
+    storeCountersInZooKeeper(false);
```
dlogothetis, line 801: nit: let's add a comment why we do this after every superstep

giraph-core/src/main/java/org/apache/giraph/worker/BspServiceWorker.java
```diff
@@ -1236,6 +1235,76 @@ public void cleanup(FinishedSuperstepStats finishedSuperstepStats)
...
+              true);
+    } catch (KeeperException.NodeExistsException e) {
+      LOG.warn("finishSuperstep: finished worker path " +
+              finishedWorkerPath + " already exists!");
+    } catch (KeeperException e) {
+      throw new IllegalStateException("Creating " + finishedWorkerPath +
+              " failed with KeeperException", e);
+    } catch (InterruptedException e) {
+      throw new IllegalStateException("Creating " + finishedWorkerPath +
+              " failed with InterruptedException", e);
```
dlogothetis: Do we want any such failures to actually fail the whole job?

giraph-core/src/main/java/org/apache/giraph/master/BspServiceMaster.java
```diff
@@ -1794,6 +1804,154 @@ private void doMasterCompute() {
...
+      }
+    }
+  }
+
+  /**
+   * Receive the counters from the workers, and aggregate them with the
+   * master counters.
+   * The aggregated counters are stored in a thrift struct
+   */
+  private void aggregateCountersFromWorkersAndMaster() {
```
dlogothetis, line 1846: There's different exception catching in here. Should we fail the job in these cases?

giraph-core/src/main/java/org/apache/giraph/counters/CustomCounters.java
```diff
@@ -0,0 +1,124 @@
...
+ * Stores the names of the counters, and the final aggregated values
+ * Since there is no way to get the list of available counters from the context,
+ * we need to call addCustomCounter to store the information about the list of
+ * available counters. This is accessed by the worker/master to fetch the
+ * final counter values based on the group and counter name.
+ * It is assumed that whenever a counter is created, addCustomCounter is called,
+ * to enable aggregations on that counter, using the Giraph mechanism for
+ * aggregating counters.
+ */
+public class CustomCounters {
```
dlogothetis, line 37: The data structures here are modified and accessed by multiple threads, they will need to be synchronized appropriately. Maybe this is why you get a concurrentmodification as per your comment in getCustomerCounters()?

giraph-core/src/main/java/org/apache/giraph/master/BspServiceMaster.java
aanchal204: I am not sure, I don't think we should be failing the jobs here. We can definitely show as a warning in the logs.

giraph-core/src/main/java/org/apache/giraph/counters/CustomCounters.java
```diff
@@ -0,0 +1,128 @@
...
+  }
+
+  /**
+   * Get the unique counter group and names
+   *
+   * @return Map of unique counter names
+   */
+  public static Set<CustomCounter> getCustomCounters() {
+    // Create a new HashSet and return that to avoid
+    // ConcurrentModificationException
```
dlogothetis: This is not relevant any more.

giraph-core/src/main/java/org/apache/giraph/master/BspServiceMaster.java
```diff
@@ -1794,6 +1808,155 @@ private void doMasterCompute() {
...
+                workerFinishedPath, true,
+                false, true);
+        LOG.info(String.format("Fetching counter values from " +
+                        "workers. Got %d out of %d",
+                workerFinishedPathList.size(), numWorkers));
+        if (workerFinishedPathList.size() == numWorkers) {
+          break;
+        }
+      } catch (KeeperException e) {
+        LOG.info("Got Keeper exception, but will retry: ", e);
```
dlogothetis: nit: make this `LOG.warn()`?

giraph-core/src/main/java/org/apache/giraph/master/input/MasterInputSplitsHandler.java
```diff
@@ -225,7 +231,16 @@ private void splitFractionReached(
    */
   public static Counter getSplitFractionDoneTimestampCounter(
       InputType inputType, double fraction, Mapper.Context context) {
-    return context.getCounter(inputType.name() + " input",
-        String.format("%.2f%% done time (ms)", fraction * 100));
+    String groupName = inputType.name() + " input";
+    String counterName = String.format("%.2f%% done time (ms)", fraction * 100);
+    Set<String> counters = COUNTER_GROUP_AND_NAMES.getOrDefault(
```
dlogothetis, line 236: Since the set of counters is known in advance (check the MasterInputSplitsHandler constructor), you can register them upon initialization. Otherwise, because this can be called from multiple threads, this map has to be synchronized too.

giraph-core/src/main/java/org/apache/giraph/master/BspServiceMaster.java
dlogothetis: Still not sure why we need a static instance here.

dlogothetis (approved):

PR #112, opened 2019-11-06
Make JobProgressTrackeClient implementation configurable

dlogothetis:
Currently, the implementation of the JobProgressTrackerClient that is used to make thrift calls to the JobProgressTrackerService is not configurable. For instance, some environments may require using a different thrift client underneath.

https://issues.apache.org/jira/browse/GIRAPH-1226

Tests:
- Ran a number of jobs with the default and with a custom client

### DIFF
--- a/giraph-core/src/main/java/org/apache/giraph/conf/GiraphConstants.java
+++ b/giraph-core/src/main/java/org/apache/giraph/conf/GiraphConstants.java
@@ -49,8 +49,10 @@ import org.apache.giraph.graph.Computation;
 import org.apache.giraph.graph.DefaultVertex;
 import org.apache.giraph.graph.DefaultVertexResolver;
 import org.apache.giraph.graph.DefaultVertexValueCombiner;
+import org.apache.giraph.graph.JobProgressTrackerClient;
 import org.apache.giraph.graph.Language;
 import org.apache.giraph.graph.MapperObserver;
+import org.apache.giraph.graph.RetryableJobProgressTrackerClient;
 import org.apache.giraph.graph.Vertex;
 import org.apache.giraph.graph.VertexResolver;
 import org.apache.giraph.graph.VertexValueCombiner;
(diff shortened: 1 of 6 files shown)


giraph-core/src/main/java/org/apache/giraph/conf/GiraphConstants.java
```diff
@@ -1201,6 +1203,13 @@
...
 
+  /** Class to use as the job progress client */
+  ClassConfOption<JobProgressTrackerClient> JOB_PROGRESS_TRACKER_CLIENT_CLASS =
+      ClassConfOption.create("giraph.jobProgressTrackerClientClass",
+        RetryableJobProgressTrackerClient.class,
+          JobProgressTrackerClient.class,
+          "Class to use to make calls to the job progress tracker service");
+
   /** Class to use to track job progress on client */
   ClassConfOption<JobProgressTrackerService> JOB_PROGRESS_TRACKER_CLASS =
```
aanchal204: As you are adding the JobProgressTrackerClient, can we rename this to JobProgressTrackerService, to be clear about the difference between the two?

giraph-core/src/main/java/org/apache/giraph/graph/RetryableJobProgressTrackerClient.java
```diff
@@ -86,6 +93,14 @@ public RetryableJobProgressTrackerClient(GiraphConfiguration conf) throws
     resetConnection();
   }
```
aanchal204, line 94: Do we need this now?

aanchal204 (commented):
Left a few comments, otherwise it looks good.

giraph-core/src/main/java/org/apache/giraph/graph/RetryableJobProgressTrackerClient.java
dlogothetis: @aanchal204, are you referring to the constructor? I left it there as we may still want to use it as we used to.

aanchal204: Alright, makes sense!

aanchal204 (approved):

PR #114, opened 2019-11-18
Upgrade Netty to 4.1.36

dlogothetis:
- Upgrades Netty to 4.1.36
- Auto-reading used to be an option. Experiments showed that disabling it not reliable with the new version, so we'll keep it permanently enabled.

https://issues.apache.org/jira/browse/GIRAPH-1228

Tests:
- Unit tests
- Ran a number of large jobs and performance is similar.

### DIFF
--- a/giraph-core/src/main/java/org/apache/giraph/comm/netty/NettyClient.java
+++ b/giraph-core/src/main/java/org/apache/giraph/comm/netty/NettyClient.java
@@ -18,6 +18,7 @@
 
 package org.apache.giraph.comm.netty;
 
+import io.netty.handler.flush.FlushConsolidationHandler;
 import org.apache.giraph.comm.flow_control.CreditBasedFlowControl;
 import org.apache.giraph.comm.flow_control.FlowControl;
 import org.apache.giraph.comm.flow_control.NoOpFlowControl;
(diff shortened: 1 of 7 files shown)


giraph-core/src/main/java/org/apache/giraph/comm/netty/NettyClient.java
```diff
@@ -335,6 +333,10 @@ protected void initChannel(SocketChannel ch) throws Exception {
             if (conf.authenticate()) {
               LOG.info("Using Netty with authentication.");
 
+              PipelineUtils.addLastWithExecutorCheck("flushConsolidation",
+                new FlushConsolidationHandler(FlushConsolidationHandler
+                  .DEFAULT_EXPLICIT_FLUSH_AFTER_FLUSHES, true),
+                handlerToUseExecutionGroup, executionGroup, ch);
```
aanchal204, line 339: Why was this needed to be added?

aanchal204 (approved):
Everything looks good to me!

giraph-core/src/main/java/org/apache/giraph/comm/netty/NettyClient.java
dlogothetis: Looking around on netty related posts, it seems this is good practice as it batches small flushes into bigger ones for efficiency.

PR #118, opened 2019-12-16
Fix issues with channel re-connection

dlogothetis:
- The LogOnErrorChannelFutureListener is called when a channel operation was complete and it was checking whether the channel failed, in which case it tried to resend any requests. Doing this required to wait until a channel had been re-established. However, doing a wait operation from the same thread that calls the handler, causes a BlockingOperationException from Netty. So this was not effective.
- I removed the call to the method that waits to re-establish the connection and send any requests. Besides, we already have a thread that periodically checks and re-sends any unsent requests, and also re-establishes any closed channels.
- Upon a channel closing, we have logic that will try to re-open the channels doing a max number of retries.. But we also had logic in the ChannelRoterator that would throw an exception if we didn't find any channel. This does not give the opportunity to re-conenct. So I removed this.
- Whenever the client closes the connection, the server catches this (Connection reset by peer) and throws an exception as well, so the job fails immediately. This does not give the opportunity to the client to re-connect. I changed this so that whenever  a server sees a "Connection reset by peer" exception, it does not fail. Still failing in all other cases.

https://issues.apache.org/jira/projects/GIRAPH/issues/GIRAPH-1230

Tests
- Snapshot tests
- Ran with job that would consistently fail due to connection errors, which now succeeds.

### DIFF
--- a/giraph-core/src/main/java/org/apache/giraph/comm/netty/ChannelRotater.java
+++ b/giraph-core/src/main/java/org/apache/giraph/comm/netty/ChannelRotater.java
@@ -25,12 +25,15 @@ import com.google.common.collect.Lists;
 import io.netty.channel.Channel;
 import io.netty.channel.ChannelFuture;
 import io.netty.channel.ChannelFutureListener;
+import org.apache.log4j.Logger;
 
 
 /**
  * Maintains multiple channels and rotates between them.  This is thread-safe.
  */
 public class ChannelRotater {
+  /** Logger */
+  private static final Logger LOG = Logger.getLogger(ChannelRotater.class);
   /** Index of last used channel */
   private int index = 0;
   /** Channel list */
@@ -73,9 +76,9 @@ public class ChannelRotater {
    */
   public synchronized Channel nextChannel() {
     if (channelList.isEmpty()) {
-      throw new IllegalArgumentException(
-          "nextChannel: No channels exist for hostname " +
-              address.getHostName());
+      LOG.warn("nextChannel: No channels exist for hostname " +
+        address.getHostName());
+      return null;
     }
 
     ++index;
--- a/giraph-core/src/main/java/org/apache/giraph/comm/netty/NettyClient.java
+++ b/giraph-core/src/main/java/org/apache/giraph/comm/netty/NettyClient.java
@@ -242,6 +242,11 @@ public class NettyClient {
   /** How many network requests were resent because connection failed */
   private final GiraphHadoopCounter networkRequestsResentForConnectionFailure;
 
+  /**
+   * Keeps track of the number of reconnect failures. Once this exceeds the
+   * value of {@link #maxConnectionFailures}, the job will fail.
+   */
+  private int reconnectFailures = 0;
 
   /**
    * Only constructor
@@ -765,9 +770,7 @@ public class NettyClient {
     Channel channel = addressChannelMap.get(remoteServer).nextChannel();
     if (channel == null) {
       LOG.warn("getNextChannel: No channel exists for " + remoteServer);
-    }
-
-    if (channel != null) {
+    } else {
       // Return this channel if it is connected
       if (channel.isActive()) {
         return channel;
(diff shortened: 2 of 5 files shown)


giraph-core/src/main/java/org/apache/giraph/comm/netty/ChannelRotater.java
```diff
@@ -73,9 +76,9 @@ public synchronized void addChannel(Channel channel) {
    */
   public synchronized Channel nextChannel() {
     if (channelList.isEmpty()) {
-      throw new IllegalArgumentException(
-          "nextChannel: No channels exist for hostname " +
-              address.getHostName());
+      LOG.warn("nextChannel: No channels exist for hostname " +
+        address.getHostName());
+      return null;
```
aanchal204, line 81: Will the job never fail in this case, then? What happens if the channel keeps getting dropped for some reason?

giraph-core/src/main/java/org/apache/giraph/comm/netty/NettyClient.java
```diff
@@ -764,25 +764,27 @@ public void operationComplete(ChannelFuture cf) {
...
-      throw new IllegalStateException(
-          "getNextChannel: No channel exists for " + remoteServer);
+      LOG.warn("getNextChannel: No channel exists for " + remoteServer);
     }
 
-    // Return this channel if it is connected
-    if (channel.isActive()) {
-      return channel;
-    }
+    if (channel != null) {
```
aanchal204: nit: can simply use else {}

giraph-core/src/main/java/org/apache/giraph/worker/BspServiceWorker.java
```diff
@@ -217,7 +217,16 @@ public BspServiceWorker(
...
     workerInfo = new WorkerInfo();
     workerServer = new NettyWorkerServer<I, V, E>(conf, this, context,
-        graphTaskManager.createUncaughtExceptionHandler());
+        graphTaskManager.createUncaughtExceptionHandler(
+          (thread, throwable) -> {
+            // If the connection was closed by the client, then we just log
+            // the error, we do not fail the job, since the client will
+            // attempt to reconnect.
+            return throwable.getMessage().startsWith(
+              "Connection reset by peer") ? false : true;
```
aanchal204: Is there a better way to determine the exception type? If not, let's make this string as a constant and use everywhere, to avoid breaking this logic.

giraph-core/src/main/java/org/apache/giraph/comm/netty/ChannelRotater.java
dlogothetis: This is called from NettyClient.getNextChannel() which will do a maximum number of retries. However, this counter is reset every time there is a failure, so it is possible that if the channel gets dropped, that we keep retrying. What I will do is make this counter global. The current max is 1000 retries, so that should be enough.

giraph-core/src/main/java/org/apache/giraph/worker/BspServiceWorker.java
dlogothetis: The exception is IOException which is very general, at this point i'd like to distinguish, and the message was the only way i could find.

As opposed to making this a constant, i'll make the whole logic a function.

giraph-core/src/main/java/org/apache/giraph/worker/BspServiceWorker.java
```diff
@@ -217,7 +219,15 @@ public BspServiceWorker(
         getGraphPartitionerFactory().createWorkerGraphPartitioner();
     workerInfo = new WorkerInfo();
     workerServer = new NettyWorkerServer<I, V, E>(conf, this, context,
-        graphTaskManager.createUncaughtExceptionHandler());
+        graphTaskManager.createUncaughtExceptionHandler(
+          (thread, throwable) -> {
+            // If the connection was closed by the client, then we just log
+            // the error, we do not fail the job, since the client will
+            // attempt to reconnect.
+            return isConnectionResetByPeer(throwable) ? false : true;
```
aanchal204: nit: 
return !isConnectionResetByPeer(throwable);

aanchal204 (approved):
Looks good!

PR #150, opened 2021-04-19
GIRAPH-1251: Add optional SSLHandler for all Netty Communication

atanu1991:
GIRAPH-1251: Add optional SSLHandler for all Netty Communication
With this change its possible to encrypt all communication via Netty
with SSL Handler. It also exposes function to add custom authn and
authz checks on tls handshake complete. Once this flag is set, all
Netty traffic will be ensured to be encrypted.

Testing:

- mvn clean verify passes.
- When the sslEncrypt flag is off there were no changes to the existing workflow.
- On enabling the flag (sslEncrypt) and having a custom
SSL Event Handler, thhe performance was varied, on some test cases
there was an improvement in perf (+25%), while few others showed
~30+% degrade. Initial tests didnt give any strong indications
of severe performance degrade throghout.

### DIFF
--- a/checkstyle.xml
+++ b/checkstyle.xml
@@ -227,7 +227,7 @@
     </module>
       <!-- Over time, we will revised this down -->
     <module name="MethodLength">
-      <property name="max" value="210"/>
+      <property name="max" value="220"/>
     </module>
     <module name="ParameterNumber">
       <property name="max" value="8"/>
--- a/giraph-core/src/main/java/org/apache/giraph/comm/netty/NettyClient.java
+++ b/giraph-core/src/main/java/org/apache/giraph/comm/netty/NettyClient.java
@@ -241,6 +241,8 @@ public class NettyClient {
   private final GiraphHadoopCounter networkRequestsResentForChannelFailure;
   /** How many network requests were resent because connection failed */
   private final GiraphHadoopCounter networkRequestsResentForConnectionFailure;
+  /** Netty SSL Handler class */
+  private final NettySSLHandler nettySSLHandler;
 
   /**
    * Keeps track of the number of reconnect failures. Once this exceeds the
@@ -325,6 +327,12 @@ public class NettyClient {
       executionGroup = null;
     }
 
+    if (conf.sslAuthenticate()) {
+      nettySSLHandler = new NettySSLHandler(true, conf);
+    } else {
+      nettySSLHandler = null;
+    }
+
     workerGroup = new NioEventLoopGroup(maxPoolSize,
         ThreadUtils.createThreadFactory(
             "netty-client-worker-%d", exceptionHandler));
@@ -395,11 +403,19 @@ public class NettyClient {
                   new SaslClientHandler(conf), handlerToUseExecutionGroup,
                   executionGroup, ch);
               PipelineUtils.addLastWithExecutorCheck("response-handler",
-                  new ResponseClientHandler(NettyClient.this, conf),
+                  new ResponseClientHandler(
+                    NettyClient.this, conf, exceptionHandler),
                   handlerToUseExecutionGroup, executionGroup, ch);
             } else {
               LOG.info("Using Netty without authentication.");
 /*end[HADOOP_NON_SECURE]*/
+
+              if (conf.sslAuthenticate()) {
+                PipelineUtils.addLastWithExecutorCheck("sslHandler",
+                  nettySSLHandler.getSslHandler(ch.alloc()),
+                  handlerToUseExecutionGroup, executionGroup, ch);
+              }
+
               PipelineUtils.addLastWithExecutorCheck("flushConsolidation",
                 new FlushConsolidationHandler(FlushConsolidationHandler
                     .DEFAULT_EXPLICIT_FLUSH_AFTER_FLUSHES, true),
@@ -429,8 +445,11 @@ public class NettyClient {
               PipelineUtils.addLastWithExecutorCheck("request-encoder",
                     new RequestEncoder(conf), handlerToUseExecutionGroup,
                   executionGroup, ch);
+              // ResponseClientHandler is the last handler in channel pipeline
+              // It handles the SSL Exception in a special way
               PipelineUtils.addLastWithExecutorCheck("response-handler",
-                  new ResponseClientHandler(NettyClient.this, conf),
+                  new ResponseClientHandler(
+                    NettyClient.this, conf, exceptionHandler),
                   handlerToUseExecutionGroup, executionGroup, ch);
 
 /*if_not[HADOOP_NON_SECURE]*/
--- a/giraph-core/src/main/java/org/apache/giraph/comm/netty/NettySSLHandler.java
+++ b/giraph-core/src/main/java/org/apache/giraph/comm/netty/NettySSLHandler.java
@@ -1,0 +1,131 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.apache.giraph.comm.netty;
+
+import io.netty.buffer.ByteBufAllocator;
+import io.netty.channel.Channel;
+import io.netty.handler.ssl.SslContext;
+import io.netty.handler.ssl.SslHandler;
+import io.netty.util.concurrent.Future;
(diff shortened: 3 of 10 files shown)


dlogothetis:
Can you add a high-level description of how the authentication mechanism works and also how this was tested?

giraph-core/src/main/java/org/apache/giraph/conf/GiraphConstants.java
```diff
@@ -1335,5 +1346,17 @@
...
+  ClassConfOption<SSLConfigReader> SSL_CONFIG_READER_CLASS =
+          ClassConfOption.create("giraph.sslConfigReader",
+              null, SSLConfigReader.class,
+                  "SSLConfigReader class - optional");
+
+  /** SSLEventHandler class - optional */
+  ClassConfOption<SSLEventHandler> SSL_EVENT_HANDLER_CLASS =
+    ClassConfOption.create("giraph.sslEventHandler",
+      null, SSLEventHandler.class,
+      "SSLEventHandler class - optional");
```
dlogothetis: Can you describe in the documentation what these are?

giraph-core/src/main/java/org/apache/giraph/comm/netty/NettySSLHandler.java
```diff
@@ -0,0 +1,102 @@
...
+import io.netty.handler.ssl.SslContext;
+import io.netty.handler.ssl.SslHandler;
+import io.netty.util.concurrent.Future;
+import org.apache.giraph.conf.ImmutableClassesGiraphConfiguration;
+import org.apache.log4j.Logger;
+
+import javax.net.ssl.SSLException;
+
+/**
+ * Utility class for all SSL related functions
```
dlogothetis: nit: this doesn't look like the typical utility class

giraph-core/src/main/java/org/apache/giraph/comm/netty/NettySSLHandler.java
```diff
@@ -0,0 +1,102 @@
...
+import io.netty.channel.Channel;
+import io.netty.handler.ssl.SslContext;
+import io.netty.handler.ssl.SslHandler;
+import io.netty.util.concurrent.Future;
+import org.apache.giraph.conf.ImmutableClassesGiraphConfiguration;
+import org.apache.log4j.Logger;
+
+import javax.net.ssl.SSLException;
+
+/**
```
dlogothetis, line 31: Can you add some description about how the handler is used, e.g. when it's created, when the methods are supposed to be called etc.

giraph-core/src/main/java/org/apache/giraph/comm/netty/SSLConfigReader.java
```diff
@@ -0,0 +1,51 @@
...
+/**
+ * An SSLConfigReader is used to handle ways in which the
+ * SSL Configs can be read. The configs include
+ * Certificate Authority file path, client and server identity file paths.
+ */
+public interface SSLConfigReader
+    extends ImmutableClassesGiraphConfigurable
+{
+  /**
+   * Read certificate authority Path from Env variable
```
dlogothetis: Regarding the naming of the methods, is the assumption that everything is is going to be read from environment variables? If not, maybe choose different naming?

giraph-core/src/main/java/org/apache/giraph/comm/netty/SSLEventHandler.java
```diff
@@ -0,0 +1,39 @@
...
+public interface SSLEventHandler
+  extends ImmutableClassesGiraphConfigurable
+{
+  /**
+   * Handle the event 'onSslHandshakeComplete' after TLS handshake
+   *
+   * @param client whether it is the client or server involved
+   * @param sslHandler the SslHandler
+   */
+  void handleOnSslHandshakeComplete(boolean client,
```
dlogothetis: nit: name the boolean `isClient`?

giraph-core/src/main/java/org/apache/giraph/comm/netty/SSLConfig.java
```diff
@@ -0,0 +1,326 @@
...
+   * Constructor to set the file paths based on verification mode
+   *
+   * @param client client or server
+   * @param verifyMode verify mode as described in the enum above
+   * @param  caPath certificate authority file path
+   * @param certPath certificate file path
+   * @param keyPath key file path
+   */
+  public SslConfig(
+      boolean client, VerifyMode verifyMode, String caPath,
```
dlogothetis: nit: name `isClient`?

giraph-core/src/main/java/org/apache/giraph/comm/netty/SSLConfig.java
```diff
@@ -0,0 +1,326 @@
...
+    private void assignCAPath()
+    {
+      this.caPath = CA_PATH.get(conf);
+
+      if (sslConfigReader != null) {
+        String envVarTlsCAPath = sslConfigReader.readCAPathFromEnv();
+        if (envVarTlsCAPath != null) {
+          this.caPath = envVarTlsCAPath;
+        }
+      }
```
dlogothetis: What is the reason for having both a Giraph option for the path and a Giraph option for the SslConfigReader class?

giraph-core/src/main/java/org/apache/giraph/comm/netty/NettySSLHandler.java
```diff
@@ -0,0 +1,102 @@
...
+   * @throws Exception
+   */
+  private void onSslHandshakeComplete(
+      Future<? super Channel> future,
+        SslHandler sslHandler) throws Exception
+  {
+    if (!future.isSuccess()) {
+      throw new SSLException("SSL Handshake failure", future.cause());
+    }
+    if (sslEventHandler != null) {
```
dlogothetis: Why would we create a NettySSLHandler, if this is not defined?

checkstyle.xml
```diff
@@ -227,7 +227,7 @@
     </module>
       <!-- Over time, we will revised this down -->
     <module name="MethodLength">
-      <property name="max" value="210"/>
+      <property name="max" value="220"/>
```
atanu1991, line 230: Change needed for maven checkstyle

giraph-core/src/main/java/org/apache/giraph/comm/netty/NettySSLHandler.java
atanu1991: Its not necessary to have sslEventHandler (onSslHandshakeComplete) event defined. Its required for facebook use case where we need to do custom authn and authz. If you want the default behavior then do nothing on onSslHandshakeComplete. So sslEventHandler is null by default

atanu1991:
> Can you add a high-level description of how the authentication mechanism works and also how this was tested?

In the summary you mean?

giraph-core/src/main/java/org/apache/giraph/comm/netty/SSLConfig.java
atanu1991: Removed this logic

giraph-core/src/main/java/org/apache/giraph/comm/netty/NettySSLHandler.java
atanu1991: Done

giraph-core/src/main/java/org/apache/giraph/comm/netty/NettyClient.java
```diff
@@ -325,6 +327,12 @@ public NettyClient(Mapper<?, ?, ?, ?>.Context context,
       executionGroup = null;
     }
 
+    if (conf.sslAuthenticate()) {
+      nettySSLHandler = new NettySSLHandler(true, conf);
+    } else {
+      nettySSLHandler = null;
+    }
```
dlogothetis, line 334: nit: this is not necessary, right?

giraph-core/src/main/java/org/apache/giraph/comm/netty/NettyServer.java
```diff
@@ -192,6 +193,12 @@ public NettyServer(ImmutableClassesGiraphConfiguration conf,
     } else {
       executionGroup = null;
     }
+
+    if (conf.sslAuthenticate()) {
+      nettySSLHandler = new NettySSLHandler(false, conf);
+    } else {
+      nettySSLHandler = null;
+    }
```
dlogothetis, line 201: same here, this is not necessayr

dlogothetis (commented):
> > Can you add a high-level description of how the authentication mechanism works and also how this was tested?
> 
> In the summary you mean?

Yes, please add in the summary a description and also a summary of how this was tested. Please do run `mvn clean verify` as well.

Also, can you comment here about the optimizations you have identified, for instance, you mentioned not creating the ssl config multiple times.

dlogothetis:
Also, have we checked whether performance is impacted?

atanu1991:
Performance will be impacted quite a bit. The testing of that is something planned for next half. For now we have to block the migration,

dlogothetis:
I agree, but I still think we should have a couple of data points. As an
extreme, say this makes runtime 3x more. Would it be acceptable?

giraph-core/src/main/java/org/apache/giraph/comm/netty/NettyClient.java
atanu1991: Compile complains :)
Final variables needs to be initalized.

giraph-core/src/main/java/org/apache/giraph/comm/netty/NettyServer.java
atanu1991: Needed as final variable

dlogothetis (approved):

PR #152, opened 2022-03-05
GIRAPH-1253: Fix vertex value reinitialized to initialValue 

siddhantbenadikar:
Summary:
1. removing reinitilizing of vertex value to initial value from first step of weighted pagerank

Test Plan:
ran a set of test jobs to check for correctness


giraph-block-app-8/src/main/java/org/apache/giraph/block_app/library/pagerank/PageRankInitializeAndNormalizeEdgesPiece.java
```diff
@@ -58,8 +54,6 @@
   public PageRankInitializeAndNormalizeEdgesPiece(
       ConsumerWithVertex<I, V, DoubleWritable, DoubleWritable> valueSetter,
```
dlogothetis: `valueSetter` should be completely removed, right?

siddhantbenadikar: yes, only reason I didn't remove this is cause it would require constructor level changes and more changes in other files in general. But I can make them and test again

dlogothetis (approved):

